PR #3, opened 2013-06-07
fixing f18 compiler error and adding ctags

lzap:
SSIA http://koji.fedoraproject.org/koji/taskinfo?taskID=5480320


domcleal:
Split into two commits and merged as 1e8f5755 and 4b537354.  Please keep maintenance bits separate from bug fixes.  Thanks @lzap!

PR #5, opened 2013-09-05
fixing comment in the release script

lzap:


domcleal:
Merged as 9a09b7c, thanks.

PR #6, opened 2013-09-05
Bump version to 1.3-develop

lzap:


domcleal:
Merged as 5dee005, thanks.  I also pushed 1.3-stable with the previous commit.

PR #11, opened 2013-11-08
Fix concurrent execution of the release script during Jenkins builds

domcleal:
Should fix the invalid SRPMS and tarballs being uploaded to koji, which were
failing with CRC errors as Jenkins runs 2/3 of these in parallel for
different dists.

Counterpart to https://github.com/theforeman/foreman-installer/pull/76

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,2 @@
 TAGS
+pkg/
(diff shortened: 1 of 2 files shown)


GregSutcliffe:
given @skottler is away, I've had a look. Seems ok from my limited knowledge of rpmbuild :)

lzap:
Updating b0837e9..b92f91d ty

PR #15, opened 2014-03-03
SELinux policies for plugins

lzap:
Do not merge yet - I will be adding policies for all our plugins in this PR.
Please review.

### DIFF
--- a/foreman.fc
+++ b/foreman.fc
@@ -40,3 +40,7 @@
 
 /opt/rh/ruby193/root/usr/share/gems/gems/passenger-.*/agents/.* -- gen_context(system_u:object_r:passenger_exec_t,s0)
 /opt/rh/ruby193/root/usr/share/gems/gems/passenger-.*/helper-scripts/prespawn -- gen_context(system_u:object_r:passenger_exec_t,s0)
+
+# Foreman Hooks plugin
+
+/usr/share/foreman/config/hooks(/.*)?   gen_context(system_u:object_r:foreman_hook_t,s0)
--- a/foreman.fc
+++ b/foreman.fc
@@ -28,6 +28,7 @@
 
 /var/run/foreman(/.*)?                  gen_context(system_u:object_r:foreman_var_run_t,s0)
 
+/usr/share/foreman/extras/noVNC/websockify\.py gen_context(system_u:object_r:websockify_exec_t,s0)
 /usr/share/foreman/script(/.*)?         gen_context(system_u:object_r:httpd_foreman_script_exec_t,s0)
 
 # Passenger non-SCL file contexts
--- a/foreman.fc
+++ b/foreman.fc
@@ -34,12 +34,13 @@
 # Passenger non-SCL file contexts
 
 /usr/lib/ruby/gems/1.8/gems/passenger-.*/agents/.* -- gen_context(system_u:object_r:passenger_exec_t,s0)
+/usr/lib64/ruby/site_ruby/1.8/x86_64-linux/agents/.* -- gen_context(system_u:object_r:passenger_exec_t,s0)
 /usr/lib/ruby/gems/1.8/gems/passenger-.*/helper-scripts/prespawn -- gen_context(system_u:object_r:passenger_exec_t,s0)
 
 # Passenger SCL file contexts
 
-/opt/rh/ruby193/root/usr/share/gems/gems/passenger-.*/agents/.* -- gen_context(system_u:object_r:passenger_exec_t,s0)
-/opt/rh/ruby193/root/usr/share/gems/gems/passenger-.*/helper-scripts/prespawn -- gen_context(system_u:object_r:passenger_exec_t,s0)
+/opt/rh/ruby193/root/usr/(share/gems/gems|lib64/gems/exts)/passenger-.*/agents/.* -- gen_context(system_u:object_r:passenger_exec_t,s0)
+/opt/rh/ruby193/root/usr/(share/gems/gems|lib64/gems/exts)/passenger-.*/helper-scripts/prespawn -- gen_context(system_u:object_r:passenger_exec_t,s0)
 
 # Foreman Hooks plugin
 
--- a/foreman.te
+++ b/foreman.te
@@ -47,6 +47,14 @@ gen_tunable(passenger_run_puppetmaster, true)
 ## </desc>
 gen_tunable(httpd_run_foreman, true)
 
+## <desc>
+## <p>
+## Determine whether passenger can connect to TCP ports
+## other than specified in the policy.
+## </p>
+## </desc>
+gen_tunable(passenger_can_connect_all, false)
+
 # define types for foreman scripts
 apache_content_template(foreman)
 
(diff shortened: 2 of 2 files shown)


lzap:
Renamed this PR to "SELinux policies for plugins".

I am done with this, @domcleal

foreman.te
```diff
@@ -217,3 +234,69 @@ optional_policy(`
...
+corenet_tcp_bind_generic_node(websockify_t)
+corenet_tcp_connect_vnc_port(websockify_t)
+corenet_tcp_bind_vnc_port(websockify_t)
+dev_read_urand(websockify_t)
+kernel_read_system_state(websockify_t)
+logging_send_syslog_msg(websockify_t)
+miscfiles_read_localization(websockify_t)
+sysnet_read_config(websockify_t)
+
+# Prevent websockify from contacting ABRT
```
domcleal: Is that to stop the exceptions from websockify from propagating into ABRT?  Bit hacky doing it here if so ;)

foreman.fc
```diff
@@ -39,3 +40,7 @@
 
 /opt/rh/ruby193/root/usr/share/gems/gems/passenger-.*/agents/.* -- gen_context(system_u:object_r:passenger_exec_t,s0)
 /opt/rh/ruby193/root/usr/share/gems/gems/passenger-.*/helper-scripts/prespawn -- gen_context(system_u:object_r:passenger_exec_t,s0)
+
+# Foreman Hooks plugin
+
+/usr/share/foreman/config/hooks.*       gen_context(system_u:object_r:foreman_hook_t,s0)
```
domcleal: Nitpick, `(/.*)?` instead of .*

domcleal:
Do we have something to permit passenger_t to execute websockify_exec_t, or does it have free reign to execute any label right now?

foreman.te
lzap: The problem here is websockify throws a python stracktrace everytime we try to connect to occupied port. And this is how Foreman works - it loops the ports until websockify starts ok. The problem is that python on RHEL6 is configured with ABRT integration by default, therefore every single try causes ABRT run, which is also quite slow.

I was not able to blacklist websockify, after some discussion with ABRT folks, I have decided this trick to mute that guy. The process is not able to connect to ABRT.

foreman.fc
lzap: I want the directory too. What is the difference here?

foreman.te
domcleal: Yeah, I'm aware... but I don't think adding SELinux policy is the way to fix that, especially as it doesn't help in non-enforcing configurations.

foreman.fc
domcleal: `(/.*)?` is usually used in SELinux fcontexts as it matches either the dir with that name and files beneath it, but not files prefixed with the same name (e.g. /usr/share/foreman/config/hooks_foo).

lzap:
It has quite loose rights on execution, yeah.

```
[root@nightly ~]# sesearch -s passenger_t -t websockify_exec_t --all
ERROR: Cannot get avrules: Neverallow rules requested but not available
Found 1 semantic av rules:
   allow passenger_t websockify_exec_t : file { read getattr execute open } ; 

Found 1 semantic te rules:
   type_transition passenger_t websockify_exec_t : process websockify_t; 
```

lzap:
But this is covered by

```
application_domain(websockify_t, websockify_exec_t)
domtrans_pattern(passenger_t, websockify_exec_t, websockify_t)
```

domcleal:
Ah yes, thanks, I missed that.

foreman.te
lzap: > Yeah, I'm aware... but I don't think adding SELinux policy is the way to fix that, especially as it doesn't help in non-enforcing configurations.

But we need to add it here anyway. We can either do dontaudit or allow.
From the two options, dontaudit makes more sense - it also solves the
problem.

ABRT features blacklisting, but that works on a package level. Therefore
on a Foreman installation, we can initiate ABRT investigation 10 times
per click.

To me `dontaudit` looks like the better option than `allow`.

Yes I admit proper fix should be done. We can either:

1) Drop BlackListedPaths in /etc/abrt/abrt-action-save-package-data.conf
(but when I initially tested this it was _not_ working). We need augeas
to do this.

2) Send upstream patch not to bubble up the main method to websockify.

I think this is a different thing, we should raise another issue/pr for
that.

## 

Later,

 Lukas "lzap" Zapletal
 irc: lzap #theforeman

foreman.fc
lzap: > `(/.*)?` is usually used in SELinux fcontexts as it matches either the dir with that name or files beneath it, but not files prefixed with the same name (e.g. /usr/share/foreman/config/hooks_foo).

Oh right got it, fixed.

## 

Later,

 Lukas "lzap" Zapletal
 irc: lzap #theforeman

domcleal:
I don't think that #5446 is our bug, it should be handled in the distribution policy.

(From my reading of the AVC, it's not the puppetmaster at all, it's just Postfix searching a dir it probably shouldn't be.)

lzap:
On Fri, Apr 25, 2014 at 05:48:11AM -0700, Dominic Cleal wrote:

> I don't think that #5446 is our bug, it should be handled in the distribution policy.

But we already carry bunch of fixes there.

## 

Later,

 Lukas "lzap" Zapletal
 irc: lzap #theforeman

foreman.te
domcleal: Ok, let's use the original issue number #4527 for investigation into that.  I'd prefer to stop it triggering the exception, frankly.

However I think if there's a macro that permits access to ABRT we should actually use this and then people can choose to log the exceptions by changing the config, particularly if we stop it triggering exceptions while finding a free port.  This will be more useful in the long term.

lzap:
Ok I have reworded issue number for the websockify, sorry. You check for this as well? Uh.

domcleal:
We don't carry any fixes for Postfix, do we - assuming as I said, that I'm reading the AVC correctly?

domcleal:
Issue number in the websockify commit still seems to point to #4527 rather than #4569?

lzap:
Youre right, I missed that tiny detail. Dropped.

lzap:
Hmm reworded incorrectly.

Thanks.

## 

Later,

 Lukas "lzap" Zapletal
 irc: lzap #theforeman

lzap:
https://github.com/kanaka/websockify/pull/126

## 

Later,

 Lukas "lzap" Zapletal
 irc: lzap #theforeman

foreman.te
domcleal: ping?

lzap:
Added one more fix

lzap:
Ok pushed, review, then I will squash.

domcleal:
:+1: thanks!

lzap:
Squashed

domcleal:
My F19 scratch build is failing: http://koji.katello.org/koji/getfile?taskID=103204&name=build.log

<pre>
Compiling targeted foreman module
foreman.te":280:ERROR 'type bin_t is not within scope' at token ';' on line 7575:
typealias bin_t alias foreman_hook_t;
/usr/bin/checkmodule:  error(s) encountered while parsing configuration
/usr/bin/checkmodule:  loading policy configuration from tmp/foreman.tmp
</pre>

foreman.te
```diff
@@ -217,3 +234,62 @@ optional_policy(`
...
+#
+
+# no rules necessary
+
+######################################
+#
+# Foreman Hooks plugin
+#
+
+typealias bin_t alias foreman_hook_t;
```
domcleal, line 280: Are we just missing bin_t from requires?

domcleal:
Merged as c446dc8..0e094fe for 1.5.0.  Thanks very much @lzap!

foreman.te
domcleal, line 280: We were, fixed.

lzap, line 280: Sorry about that, thanks.

I am currently enjoying quite interesting gdb session :-)

## 

Later,

 Lukas "lzap" Zapletal
 irc: lzap #theforeman

PR #16, opened 2014-04-30
fixes #5487 - fixed paths in selinux-relabel script

lzap:


domcleal:
Looks fine, will wait on Ade's response on the ticket though.

domcleal:
Merged as b5f521e for Foreman 1.5.0, thanks @lzap!

PR #17, opened 2014-05-22
Fixes #5882 - allow Foreman app to reach dynflow socket

lzap:


foreman.te
```diff
@@ -213,6 +218,10 @@ optional_policy(`
         # Fedora.
         allow httpd_t puppet_etc_t:dir getattr;
         allow httpd_t puppet_port_t:tcp_socket name_connect;
+        # Puppet master does touch xattr which SELinux sees as relabelto
```
domcleal, line 221: This appears unrelated to dynflow...

foreman.te
```diff
@@ -168,6 +168,11 @@ optional_policy(`
         # Allow Foreman to write to the SQlite databases
         read_files_pattern(passenger_t, foreman_db_t, foreman_db_t)
         write_files_pattern(passenger_t, foreman_db_t, foreman_db_t)
+
+        # Allow Foreman to reach Dynflow engine
+        allow passenger_t foreman_var_run_t:sock_file write;
+        init_stream_connect_script(passenger_t)
+        unconfined_stream_connect(passenger_t)
```
domcleal, line 175: That sounds fairly open?

domcleal:
What AVCs are generated?  The bug report doesn't say.

foreman.te
```diff
@@ -168,6 +168,11 @@ optional_policy(`
         # Allow Foreman to write to the SQlite databases
         read_files_pattern(passenger_t, foreman_db_t, foreman_db_t)
         write_files_pattern(passenger_t, foreman_db_t, foreman_db_t)
+
+        # Allow Foreman to reach Dynflow engine
```
domcleal, line 172: If this is a plugin, then it should probably be in that section.

foreman.te
lzap, line 221: Yes, but I have the denials in the BZ. You you want me to split the upstream?

lzap:
https://bugzilla.redhat.com/show_bug.cgi?id=1098244

foreman.te
domcleal, line 221: Yes.

foreman.te
lzap, line 175: Hmm yes I am out of ideas how to prevent Foreman from failing in enforcing when dynflow (unconfined) is connecting to Foreman via sockets.

domcleal:
Ok, but please keep our redmine tickets up to date and relevant, they're what we use in Foreman, not bugzilla.  Use Bryan's clone script if you need to.

foreman.te
lzap, line 172: Well I postponed delivering selinux policy for tasks because I spent whole day wrinting that and testing that and the result is not yet ready. I am not sure I can deliver that by end of friday. Daemons rubygem is the root cause of this.

foreman.te
lzap, line 175: This is really pain, I need to write the policy for foreman-tasks really. This is not a solution, permissive is better than merging this.

lzap:
Ok closing this one, I need to finish my proper solution to that.

PR #18, opened 2014-05-23
Foreman-tasks selinux policy

lzap:
Note that I also fix three AVCs generated by puppet master. I did not
investigate this further. I can split into two separate commits if needed, they
both have been reported in one issue.

### DIFF
--- a/foreman.fc
+++ b/foreman.fc
@@ -45,3 +45,8 @@
 # Foreman Hooks plugin
 
 /usr/share/foreman/config/hooks(/.*)?   gen_context(system_u:object_r:foreman_hook_t,s0)
+
+# Foreman Tasks plugin
+
+/usr/share/gems/gems/foreman-tasks-*/bin/foreman-tasks -- gen_context(system_u:object_r:foreman_tasks_exec_t,s0)
+/opt/rh/ruby193/root/usr/share/gems/gems/foreman-tasks-.*/bin/foreman-tasks -- gen_context(system_u:object_r:foreman_tasks_exec_t,s0)
--- a/foreman.te
+++ b/foreman.te
@@ -293,3 +293,12 @@ typealias bin_t alias foreman_hook_t;
 #
 
 # no rules necessary
+
+######################################
+#
+# Foreman Tasks plugin
+#
+
+# the plugin daemon uses daemon gem for the backround job
+type foreman_tasks_exec_t;
+init_daemon_domain(passenger_t, foreman_tasks_exec_t)
--- a/foreman.te
+++ b/foreman.te
@@ -155,11 +155,9 @@ dev_read_rand(passenger_t)
 
 optional_policy(`
     tunable_policy(`passenger_run_foreman', `
-        read_files_pattern(httpd_t, foreman_lib_t, foreman_lib_t)
-        manage_files_pattern(passenger_t, foreman_lib_t, foreman_lib_t)
-        manage_dirs_pattern(passenger_t, foreman_lib_t, foreman_lib_t)
-        manage_files_pattern(passenger_t, foreman_var_run_t, foreman_var_run_t)
-        manage_dirs_pattern(passenger_t, foreman_var_run_t, foreman_var_run_t)
+        admin_pattern(httpd_t, foreman_lib_t, foreman_lib_t)
+        admin_pattern(passenger_t, foreman_lib_t, foreman_lib_t)
+        admin_pattern(passenger_t, foreman_var_run_t, foreman_var_run_t)
 
         # Allow Foreman to connect to hosts and guests
         corenet_tcp_connect_virt_port(passenger_t)
--- a/foreman.te
+++ b/foreman.te
@@ -81,6 +81,9 @@ logging_log_file(foreman_log_t)
 type foreman_var_run_t;
 files_pid_file(foreman_var_run_t)
 
+type foreman_proxy_port_t;
+corenet_port(foreman_proxy_port_t)
+
 require{
     type bin_t;
     type httpd_t;
@@ -97,6 +100,7 @@ require{
     type puppetmaster_exec_t;
     type puppetmaster_t;
     type sysctl_net_t;
+    type websm_port_t;
 }
 
 #######################################
@@ -127,6 +131,12 @@ allow passenger_t self:process signull;
 allow passenger_t self:tcp_socket listen;
 
 miscfiles_read_localization(passenger_t)
+
+# Allow Foreman (Satellite) to connect to Foreman Proxy (Capsule) on port 9090
+allow passenger_t websm_port_t:tcp_socket name_connect;
+
+# Allow Foreman to connect to Foreman Proxy on a defined port
+allow passenger_t foreman_proxy_port_t:tcp_socket name_connect;
 
 # Allow Foreman to connect to PostgreSQL
 corenet_tcp_connect_postgresql_port(passenger_t)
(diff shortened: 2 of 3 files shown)


lzap:
Just to let you know - the three denials are out of scope for this PR because it turns out that Apache2 was started incorrectly, thus there is a mismatch between unconfied_u and system_u. Normally, you should not see this error.

foreman.te
```diff
@@ -293,3 +293,18 @@ typealias bin_t alias foreman_hook_t;
...
+#
+# Foreman Tasks plugin
+#
+
+# the plugin daemon uses daemon gem for the backround job
+type foreman_tasks_exec_t;
+init_daemon_domain(passenger_t, foreman_tasks_exec_t)
+
+# the daemon gem has monitor process to respawn worker job
+allow passenger_t initrc_t:process { signal signull };
```
domcleal: What's in the initrc_t domain?

lzap: This is likely from the foreman-tasks init5 script.

domcleal:
Can you give an overview of how foreman-tasks is deployed?  (Processes, what's spawning what, etc.)

foreman.te
lzap: Yes, there is a killproc.

lzap:
There is /usr/bin/foreman-tasks which is here started from this script:

https://github.com/theforeman/foreman-tasks/blob/master/deploy/foreman-tasks.init

It spawns two processes

```
unconfined_u:system_r:passenger_t:s0 foreman 15092 0.0  0.7 928844 142988 ?    Sl   05:01   0:15 dynflow_executor                                  
unconfined_u:system_r:passenger_t:s0 foreman 15095 0.0  0.3 311924 61336 ?     Sl   05:01   0:00 dynflow_executor_monitor 
```

They talk to the Foreman core via the socket. @iNecas did I miss something?

domcleal:
What do they do, is there any reason they're in passenger_t instead of their own domain?  Do they load the Rails application again?

foreman.te
domcleal: Wouldn't that be killing passenger_t domain processes, not initrc_t?

lzap:
Yes, Its essentialy another Foreman.

lzap:
I need to run today, merge it will. Its tested, I'd like to have it since you are on PTO on Monday.

lzap:
@domcleal so I am putting more issues in this PR - see list of commits.

I fixed symlink reading from /usr/share and also loosed rules for Rails a bit which now also includes rules to create/read sockets, therefore no extra rule is necessary in the foreman-tasks section.

foreman.te
lzap: Oh of course, I am removing this one, because when I was restarting in for the first time, the instance was running in initrc_t!

Good catch.

lzap:
Rebased, this PR now cointains 6 issues fixed.

foreman.te
```diff
@@ -128,6 +129,12 @@ allow passenger_t self:tcp_socket listen;
 
 miscfiles_read_localization(passenger_t)
 
+# Allow Foreman to connect to Foreman Proxy on port 9090 (Satellite)
```
domcleal: Katello "capsule"

domcleal:
This is harder to review when they're all in one PR, I'd prefer them separately in future.  Are you missing a commit - there are only five, not six?
1. foreman-tasks looks fine
2. symlinks looks fine
3. I don't follow the analysis in the bug report, as Puppet should set the context abiding by the fcontext list, and failing that we can force it from the installer.  It's possible that Puppet module content is labelled as puppet_etc_t, so I'd prefer to keep this type.
4. puppetmaster is fine
5. proxy port is fine, subject to comment

domcleal:
Can we add default ports to foreman_proxy_port_t anywhere?  (8443, 8000)

Edit: oh, 8443's already handled by the web ports, I remember you saying.  Is 8000 (the new HTTP-only port in the proxy modularisation PR) included?

lzap:
> This is harder to review when they're all in one PR, I'd prefer them separately in future.  Are you missing a commit - there are only five, not six?

Yeah ok. My fault - I miscounted them - they are five. I have one
another locally, will push into different one then.

The problem I am trying to solve with one PR is amount of git conflicts
I'd have because this file is small and I usually change sections near
each other -> conflicts.

> 1. I don't follow the analysis in the bug report, as Puppet should set the context abiding by the fcontext list, and failing that we can force it from the installer.  It's possible that Puppet module content is labelled as puppet_etc_t, so I'd prefer to keep this type.

But is there any reason for keeping this? I think the special type was
unnecessary in the first place. Also, we are not able to create files
with different types within a directory of different type (well there is
one way but I'd rather not use it).

Anyway one important note - if we want to run the ENC script in a
different domain than passenger_t, then we'd need to set a context for
it anyway. I am not sure if this is worth the effort.

## 

Later,

 Lukas "lzap" Zapletal
 irc: lzap #theforeman

lzap:
> Can we add default ports to foreman_proxy_port_t anywhere?  (8443, 8000)

We can't because SELinux won't allow to redefine existing ports. These
are already defined in http_port_t.

## 

Later,

 Lukas "lzap" Zapletal
 irc: lzap #theforeman

lzap:
Damn, can you write which bug do you do not understand? :-)

## 

Later,

 Lukas "lzap" Zapletal
 irc: lzap #theforeman

domcleal:
My comments corresponded to the commit order as shown in the PR, so bug #5924.

I think keeping the binary (node.rb) labelled differently to user config and content (puppet_etc_t) so we don't enable execution of puppet_etc_t content is a good reason for having a separate label... and that we've already gone to this effort.

lzap:
Right, I understand. Now how do you recommend we should solve this? We can instruct our puppet to drop the file with particular selinux context I think, but we need to make sure this works with debian too. Is this what you prefer?

domcleal:
Yeah, I do, though am unsure why it's not already working that way to be honest as Puppet should call matchpathcon (http://docs.puppetlabs.com/references/stable/type.html#file-attribute-selinux_ignore_defaults), would be nice to know.  If you specify seltype then it should ignore it on OSes or Puppet runs where that property can't be managed (as the SELinux lib would be missing), but would be good to test.

lzap:
Ok I will include seltype then.

lzap:
Pushed, I think we are done over here.

foreman.fc
```diff
@@ -45,3 +45,7 @@
 # Foreman Hooks plugin
 
 /usr/share/foreman/config/hooks(/.*)?   gen_context(system_u:object_r:foreman_hook_t,s0)
+
+# Foreman Tasks plugin
+
+/opt/rh/ruby193/root/usr/share/gems/gems/foreman-tasks-.*/bin/foreman-tasks -- gen_context(system_u:object_r:foreman_tasks_exec_t,s0)
```
domcleal, line 52: Missing the Fedora path too: /usr/share/gems/gems/foreman-tasks-0.5.3/bin/foreman-tasks

Any idea what this is, and if it needs labelling? /usr/share/gems/gems/foreman-tasks-0.5.3/bin/dynflow-executor

domcleal:
Yep, one comment, otherwise looks fine.  One of the commit messages says "Fixed" instead of "Fixes" btw.

foreman.fc
lzap, line 52: Oh fixed, good catch.

@iNecas informed me that the other binary will go away in the next build, it is unused. I am not labelling it then.

Thanks

domcleal:
Merged as 9554703, fcc3110, a39d8de, b13ec51.  Thanks @lzap!

PR #21, opened 2014-06-03
Katello selinux policy and other fixes

lzap:
After week of Satellite 6 testing, it looks like we only miss those rules.

I was not able to find out what is spawning the `consoletype` binary. Searched
most of our plugins, Foreman, Katello, tog, passenger itself. But it looks
harmless - only returns console type, it has it's own domain which is safe
enough.

There is still one bug - something binds _random_ udp port. Open topic.

### DIFF
--- a/foreman-selinux-enable
+++ b/foreman-selinux-enable
@@ -1,11 +1,16 @@
 #!/bin/bash
 
-# install policy modules in one transaction
 for selinuxvariant in targeted
 do
-  /usr/sbin/semodule -s $selinuxvariant -l >/dev/null 2>&1 && \
-  /usr/sbin/semanage -S $selinuxvariant -i - << _EOF
-  module -a /usr/share/selinux/${selinuxvariant}/foreman.pp.bz2
-  boolean -m --on httpd_setrlimit
-_EOF
+  if /usr/sbin/semodule -s $selinuxvariant -l >/dev/null; then
+    # TRANSACTION 1
+    /usr/sbin/semanage -S $selinuxvariant -i - << _EOT1
+      module -a /usr/share/selinux/${selinuxvariant}/foreman.pp.bz2
+      boolean -m --on httpd_setrlimit
+_EOT1
+    # TRANSACTION 2
+    /usr/sbin/semanage -S $selinuxvariant -i - << _EOT2
+      port -a -t elasticsearch_port_t -p tcp 9200-9300
+_EOT2
+  fi
 done
--- a/foreman.te
+++ b/foreman.te
@@ -267,10 +267,11 @@ require {
     type vnc_port_t;
 }
 
-allow websockify_t self:netlink_route_socket { bind create getattr nlmsg_read };
-allow websockify_t self:tcp_socket { setopt bind create listen accept connect shutdown };
-allow websockify_t self:udp_socket { getattr ioctl create connect };
+allow websockify_t self:netlink_route_socket { all_netlink_route_socket_perms };
+allow websockify_t self:tcp_socket { all_tcp_socket_perms };
+allow websockify_t self:udp_socket { all_udp_socket_perms };
 
+apache_search_config(websockify_t)
 corenet_tcp_bind_generic_node(websockify_t)
 corenet_tcp_connect_vnc_port(websockify_t)
 corenet_tcp_bind_vnc_port(websockify_t)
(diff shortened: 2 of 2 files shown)


foreman.te
```diff
@@ -129,6 +129,7 @@ miscfiles_read_localization(httpd_foreman_script_t)
 allow passenger_t self:capability sys_resource;
 allow passenger_t self:process signull;
 allow passenger_t self:tcp_socket listen;
+allow passenger_t self:process execmem;
```
domcleal: What's this from?  It's not a great thing to be permitting.

foreman-selinux-enable
```diff
@@ -1,11 +1,20 @@
...
-_EOF
+  if /usr/sbin/semodule -s $selinuxvariant -l >/dev/null; then
+    # TRANSACTION 1
+    /usr/sbin/semanage -S $selinuxvariant -i - << _EOT1
+      module -a /usr/share/selinux/${selinuxvariant}/foreman.pp.bz2
+      boolean -m --on httpd_setrlimit
+_EOT1
+    # TRANSACTION 2
+    /usr/sbin/semanage -S $selinuxvariant -i - << _EOT2
+      port -a -t elasticsearch_port_t -p tcp 9200-9300
```
domcleal, line 13: Can't this be in the policy?  network_port or portcon?

foreman.te
lzap: It's not, do you have any other idea what I can do instead? It must be something in the passenger itself, I doubt Foreman or Katello is responsible for this.

foreman-selinux-enable
lzap, line 13: What can be in the policy? I dont follow.

domcleal, line 13: The assigning of ports to elasticsearch_port_t.

foreman.te
domcleal: Why's it only coming up now?  I thought you were adding policy changes for Katello.

lzap: > Why's it only coming up now?  I thought you were adding policy changes for Katello.

Perhaps passenger is under heavier load with Katello services or
something. I have no clue, really. Do you think we should ignore the
denial for now and only add it when we will have more of these reported?

## 

Later,

 Lukas "lzap" Zapletal
 irc: lzap #theforeman

foreman-selinux-enable
lzap, line 13: > > -      port -a -t elasticsearch_port_t -p tcp 9200-9300
> 
> The assigning of ports to elasticsearch_port_t.

Please teach me how to do that. I am only aware of this method of adding
ports.

## 

Later,

 Lukas "lzap" Zapletal
 irc: lzap #theforeman

lzap, line 13: Ping?

domcleal, line 13: I think something like `network_port(elasticsearch, tcp,9200-9300,s0)` (looking at selinux-policy), but very much untested :)

foreman.te
domcleal: I'd prefer to ignore it for now and see under what circumstances it comes up, I find it quite concerning.

lzap: All done. Tested.

## 

Later,

 Lukas "lzap" Zapletal
 irc: lzap #theforeman

lzap: I can even delete port which is defined this way.

## 

Later,

 Lukas "lzap" Zapletal
 irc: lzap #theforeman

lzap:
Pushed, tested, confirmed with @mgrepl that this is allowed.

lzap:
Renamed this pull request, will stash other selinux commits in this one, I have few to do.

lzap:
Strange, when I want to recompile the katello policy patch today, I am getting error. It was working the other day (this is a clean RHEL6 host).

```
Compiling targeted foreman module
/usr/bin/checkmodule:  loading policy configuration from tmp/foreman.tmp
foreman.te":294:ERROR 'syntax error' at token 'portcon' on line 12688:
portcon tcp 9200-9300 system_u:object_r:elasticsearch_port_t:s0
# We use this instead of port -a -t elasticsearch_port_t -p tcp 9200-9300
/usr/bin/checkmodule:  error(s) encountered while parsing configuration
```

lzap:
Ok I amended the change, I am adding the port using semanage again, portcon is not working right. I am not able to tell why.

Pushed other fix to this PR - websockify dontaudit fix.

lzap:
Dom if you can merge, please do. Otherwise I will start pushing other fixes into this one. I have two dozens of bugs.

lzap:
Ok it looks like I won't add more commits here today, @domcleal can you do the final review please and merge if it is fine? Thanks!

domcleal:
Merged as 0578ccf and 93006b8 for Foreman 1.5.2 and 1.6.0 respectively.  Thanks @lzap.

PR #23, opened 2014-06-24
fixes #6360 - enable pulp_manage_puppet when possible

lzap:


lzap:
The relevant pulp patch is here: https://github.com/pulp/pulp/pull/1020

domcleal:
This is configuration of policy, so shouldn't be here, put it in the Katello installer or something.

lzap:
Right.

lzap:
https://github.com/Katello/puppet-pulp/pull/20

PR #24, opened 2014-07-25
Fixes #6780 - Remove elasticsearch port on uninstall

lzap:


lhh:
These changes look correct.

lzap:
@domcleal what do you think?

foreman-selinux-disable
```diff
@@ -1,8 +1,16 @@
...
+# a separate transaction. Do not forget to edit countepart file
+# (enable/disable) when updating this script.
 for selinuxvariant in targeted
 do
-  /usr/sbin/semodule -s $selinuxvariant -l > /dev/null 2>&1 \
-    && /usr/sbin/semodule -s $selinuxvariant -r foreman
+  if /usr/sbin/semodule -s $selinuxvariant -l >/dev/null; then
+    # Set booleans
+    /usr/sbin/semanage -S $selinuxvariant -i - << ____EOT
+      port -d -t elasticsearch_port_t -p tcp 9200-9300
```
domcleal: I think this will break if the ports are re-configured?  e.g.

<pre>
[root@foreman-el6 ~]# semanage -S targeted -i -
port -d -t elasticsearch_port_t -p tcp 100-200
/usr/sbin/semanage: Port tcp/100-200 is not defined
</pre>


I'm still very unsure that this should ever have been in here, I think the policy was the right place.

foreman-selinux-disable
```diff
@@ -1,8 +1,16 @@
...
-# uninstall policy modules
+# Unload foreman policy and set booleans. Dependant booleans must be managed in
+# a separate transaction. Do not forget to edit countepart file
+# (enable/disable) when updating this script.
 for selinuxvariant in targeted
 do
-  /usr/sbin/semodule -s $selinuxvariant -l > /dev/null 2>&1 \
-    && /usr/sbin/semodule -s $selinuxvariant -r foreman
+  if /usr/sbin/semodule -s $selinuxvariant -l >/dev/null; then
+    # Set booleans
```
domcleal: Inaccurate

foreman-selinux-disable
lzap: You can reconfigure them easily. I think you issued wrong command `-d` for delete. What you want is perhaps:

```
semanage port -d -t elasticsearch_port_t -p tcp 9200-9300
semanage port -a -t elasticsearch_port_t -p tcp 11200-11300
```

The policy hardcodes the port and it was wrong idea from the day one. Please bring some arguments why we should not define ports this way.

I agree this should be present in elasticsearch policy, but since there is not any (and most likely we'd need to start one), let's carry it until we have resources to maintain it. But that's another story.

domcleal: Sorry, the example was thinking about what would happen if we called this disable snippet when the port had already been changed using the commands you say.  If a user customised it to 11200-11300, then the disable script ran (using my example command, s/100-200/9200-9300/) then I think it would fail and produce an error during uninstall still.

I don't mean that it should be in elasticsearch policy (we know that), but I mean that it should be in foreman.te ideally, rather than being implemented in enable/disable scripts.

foreman-selinux-disable
lzap: You mean to fix the comment, right?

Once I fix my messed git repo... :-D

domcleal: yup

foreman-selinux-disable
lzap: I've just confirmed (again) with @mgrepl that selinux module is not allowed to define ports in the policy (.te) file. Only core policy can do that. It looks like this is the only option.

I found some -C and -E extra arguments which will make the uninstall snippet short.

lzap:
Fixed, try now.

foreman-selinux-disable
domcleal: Ah, I didn't know core/non-core policy was the issue we had, thanks for checking.

domcleal:
Merged as ae6f1a6 for Foreman 1.6, thanks @lzap!

PR #29, opened 2014-08-21
Fixes #7193 - fix for RHEL 7.0 katello installer

lzap:
Fix for 7.0 until fix is delivered for 7.1 (or errata):

https://bugzilla.redhat.com/show_bug.cgi?id=1130086

### DIFF
--- a/foreman.te
+++ b/foreman.te
@@ -308,6 +308,16 @@ ifndef(`distro_rhel7', `
     consoletype_exec(passenger_t)
 ')
 
+# fix for RHEL 7.0 - https://bugzilla.redhat.com/show_bug.cgi?id=1130086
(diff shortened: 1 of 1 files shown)


domcleal:
I don't think we should copy core policy, particularly maintaining it for a package that doesn't exist in upstream.  I'd recommend requesting a z-stream for that fix if it's important.

lzap:
I need to add it downstream anyway, we need to carry on with our testing process. Branching.

lzap:
One idea - since I have work in progress branch of splitting everything into their own packages, how a new package called foreman-selinux-katello-workarounds sounds like? I could be able to put those rules there (we would not ship it upstream).

PR #30, opened 2014-08-21
Fixes #7198 - allowed httpd_t to read/write to passenger socket

lzap:
SSIA

If I can have quick look on both patches.

### DIFF
--- a/foreman.te
+++ b/foreman.te
@@ -191,7 +191,7 @@ optional_policy(`
     tunable_policy(`passenger_run_foreman', `
         allow passenger_t self:process getsession;
         fs_rw_anon_inodefs_files(passenger_t)
-        allow passenger_t httpd_t:unix_stream_socket getattr;
+        allow passenger_t httpd_t:unix_stream_socket { read write getattr };
         fs_getattr_xattr_fs(passenger_t)
         ifdef(`distro_rhel7', `
             allow passenger_t self:capability2 block_suspend;


domcleal:
I can't reproduce this on EL7, nor can I find a foreman-debug archive associated with the bug.  What's the socket?

lzap:
I pasted wrong AVC, it's this one:

```
----
time->Wed Aug 20 17:59:40 2014
type=SYSCALL msg=audit(1408571980.864:397): arch=c000003e syscall=0 success=yes exit=1 a0=3 a1=7fff5d302860 a2=1 a3=0 items=0 ppid=19265 pid=19297 auid=4294967295 uid=0 gid=0 euid=0 suid=0 fsuid=0 egid=0 sgid=0 fsgid=0 tty=(none) ses=4294967295 comm="PassengerWatchd" exe="/usr/lib64/gems/ruby/passenger-4.0.18/agents/PassengerWatchdog" subj=system_u:system_r:passenger_t:s0 key=(null)
type=AVC msg=audit(1408571980.864:397): avc:  denied  { read } for  pid=19297 comm="PassengerWatchd" path="socket:[94029]" dev="sockfs" ino=94029 scontext=system_u:system_r:passenger_t:s0 tcontext=system_u:system_r:httpd_t:s0 tclass=unix_stream_socket
----
```

lzap:
Oh right, I should say - this happens when you start foreman-tasks which boots up foreman environment and creates a socket file:

```
[root@lzap-sat6-snap6 ~]# ll /usr/share/foreman/tmp/sockets/
total 0
srw-------. 1 foreman foreman 0 Aug 21 10:30 dynflow_socket
```

Maybe I should move it to the Foreman Tasks section (which I am going to break into separate package soon - on it).

lzap:
Sorry I was on the wrong path with foreman-tasks. The reproducer is to install latest drop and during puppet when forman-tasks is being started passenger does this. Although foreman-tasks run in the same domain, it is really the passenger (cos foreman-tasks has nothing to do with httpd_t).

I did not mention that this has dontaudit rule, so you can't see it. Passenger simply does not start without any reason given. If you do permissive, ti works. If you switch back to enforcing, it does not. If you allow this, it works fine.

To see this disable dontautit rules.

lzap:
Unfotunately I have no idea how to trigger this with Foreman Core on RHEL7. I have asked the SELinux team to explain why they added this dontaudit (this is not our rule).

lzap:
Ok its in the base policy "to avoid leaks" I've been told.

iNecas:
I can confirm that passenger doesn't start when this policy is missing, and we need to read/write from/to the socket for the web server <-> dynflow executor communication

iNecas:
:+1:

domcleal:
Merged as 0a4d60f for Foreman 1.6.0, thanks @lzap.

PR #31, opened 2014-09-04
Fixes #7346 - Added OpenStack port 5000 via boolean

lzap:


domcleal:
Merged as 24a501b, thanks @lzap!

PR #28, opened 2014-08-20
Fixes #7178 - allowed passenger_t to execmem

lzap:
Tested, see the ticket.

PR #32, opened 2014-09-26
Refs #7178 - removed passenger_t execmem rule

lzap:
This reverts commit d867377e56451fc43030a30958499d34e6f4e485.

Introduced here: https://github.com/theforeman/foreman-selinux/pull/28

Not required because: https://github.com/Katello/katello/pull/4673


domcleal:
Thanks!  Merged as e86bf00.

PR #34, opened 2014-09-27
fixes #7729 - allow websockify to read certs

stbenjam:
This is definitely needed for Katello because our certs are in `/etc/pki`.

### DIFF
--- a/foreman.te
+++ b/foreman.te
@@ -331,6 +331,9 @@ ifndef(`distro_rhel7', `
 # Katello does connect to Elasticsearch services
 allow passenger_t elasticsearch_port_t:tcp_socket name_connect;
 
+# Katello uses certs in /etc/pki/katello for websockets
+miscfiles_read_certs(websockify_t)
+
 ######################################
 #
 # Foreman Bootdisk plugin


lzap:
Hey, this looks like a different bux than 7719. Can you file it with the denial and include some info about why we need this? Also, note that we have a Katello plugin section in the .te file - please put your rules there. Thanks!

stbenjam:
@lzap I think this is the cause of 7719, the SSH denials he was seeing were unrelated I think and they need a separate issue if they can be reproduced.

lzap:
Right, then let's paste the correct denials in the ticket. I haven't seen them yet :-)

stbenjam:
@lzap Updated

domcleal:
s/refs/fixes/ unless there's something else required

lzap:
This seems legit to me, out of my curiousity, where do you setup this? In the installer?

stbenjam:
Just in the Katello default answer file, we set the certs to the Katello ones instead of the Puppet-generated ones:

  https://github.com/Katello/katello-installer/blob/master/config/answers.katello-installer.yaml#L20

lzap:
@domcleal I dont see "websockets" in our installer, so we don't support that right? Just checking.

stbenjam:
@lzap It's there by default in Foreman

https://github.com/theforeman/puppet-foreman/blob/master/manifests/params.pp#L202

stbenjam:
You use Puppet certs though, which is allowed by https://github.com/theforeman/foreman-selinux/blob/develop/foreman.te#L305.

lzap:
Thanks merged as 01ba3e1 !

PR #36, opened 2014-10-09
Fixes #7719 - added ssh rules for libvirt CR

lzap:


lzap:
I noticed during reproducing this bug that I have the denials in my audit log, but Foreman does work without it for me. It relly looks like my box was in `Permissive` but it was not. User reports that Foreman was prevented in his case.

This patch allows the execution. Also, I want to create two patches for foreman and foreman-packaging to deploy the `/usr/share/foreman/.ssh` directory so it is properly labeled in all instalations so users will not struggle with ssh setup anymore. Please test thanks.

lzap:
Ok here is the patch: https://github.com/theforeman/foreman/pull/1842

lzap:
Packaging done, tested on RHEL6. Ready for review :-)

domcleal:
Should this be inside a boolean, like with OpenStack?

lzap:
You mean the whole libvirt-ssh part?

lzap:
Note that I've added couple of new rules which were preventing the console to start.

lzap:
I introduced a boolean. If you insist, I can create new PR with the console fix, which seemed to be one issue. Or we can merge it together.

lzap:
Ok its now split into two commits, I think this is reasonable, what you think? Got a min to test all my selinux pull requests? ;-)

domcleal:
Merged as b696bcf, 9fb8417 for Foreman 1.6.3.  Thanks @lzap!

PR #37, opened 2014-10-14
Fixes #7932 - added LDAP support for passenger

lzap:


domcleal:
Merged as aab37c6 for Foreman 1.6.3, thanks @lzap.

PR #38, opened 2014-11-03
Adding satellite wrappers

lzap:
I'd like to keep more downstream fixes upstream. Currently we need to allow
9090 port in Capsule, but I expect more of these. This is just a POC, do not
merge.

If you like it I will go ahead with putting some content in and creating a
ticket.


domcleal:
:-1:

PR #39, opened 2014-11-03
Fixes #8030 - allow UDP port binds for ruby net package

lzap:
I've confirmed with SELinux team.


lzap:
FYI the reason for that randomization is a weakness in DNS protocol: http://linuxsysadminblog.com/2008/12/how-to-check-if-your-dns-server-implements-source-port-randomization/ and https://www.dns-oarc.net/oarc/services/porttest

domcleal:
Merged as d35fada, thanks @lzap.

PR #40, opened 2014-11-04
Fixes #8241 - added rules for local sockets libvirt connections

lzap:


domcleal:
Merged as e2863e4, thanks @lzap.

PR #41, opened 2014-12-19
Fixes #8683 - allowed Celery port for Katello

lzap:
Not introducing boolean for this one, I want to extract all the Katello bit
into separate module post 6.1.

### DIFF
--- a/foreman.te
+++ b/foreman.te
@@ -393,6 +393,13 @@ allow passenger_t elasticsearch_port_t:tcp_socket name_connect;
 # Katello uses certs in /etc/pki/katello for websockets
 miscfiles_read_certs(websockify_t)
 
+# Katello installer configures Pulp to listen on port 5000
+ifdef(`distro_rhel6', `
+    corenet_tcp_bind_commplex_port(passenger_t)
+',`
+    corenet_tcp_bind_commplex_main_port(passenger_t)
+')
+
 ######################################
 #
 # Foreman Bootdisk plugin


foreman.te
```diff
@@ -393,6 +393,14 @@ allow passenger_t elasticsearch_port_t:tcp_socket name_connect;
 # Katello uses certs in /etc/pki/katello for websockets
 miscfiles_read_certs(websockify_t)
 
+# Katello installer configures Pulp to listen on port 5000
+ifdef(`distro_rhel6', `
+    corenet_tcp_bind_commplex_port(passenger_t)
```
jlsherrill, line 398: what about rhel 7?

domcleal, line 398: line 400?

lzap, line 398: Yeah this is the cleanest way to do this IMHO. Modern Fedoras + RHEL7 are "the default".

lzap:
I removed the empty line, does it work for you on both RHEL6 and 7 Justin?

anthonyclarka2:
lzap, I tried a clean install on centos 7 with the changed /usr/share/doc/foreman-selinux-1.7.1/foreman.te file but still couldn't bind to port 5000.

Once I put selinux in permissive mode, it worked fine.

EDIT:  at least it could bind to 5000, it found another error.

jlsherrill:
@lzap sorry for the delay.  I tested this on el7 with this scratch build:  http://koji.katello.org/koji/taskinfo?taskID=214578

and still see the problem:

```
type=AVC msg=audit(1422037912.969:833): avc:  denied  { name_bind } for  pid=978 comm="httpd" src=5000 scontext=system_u:system_r:httpd_t:s0 tcontext=system_u:object_r:commplex_main_port_t:s0 tclass=tcp_socket
type=AVC msg=audit(1422037912.970:834): avc:  denied  { name_bind } for  pid=978 comm="httpd" src=5000 scontext=system_u:system_r:httpd_t:s0 tcontext=system_u:object_r:commplex_main_port_t:s0 tclass=tcp_socket
```

jlsherrill:
Also tested using httpd_t instead of passenger_t and saw that appeared to have no affect either.

lzap:
@domcleal mind merging?

domcleal:
@lzap I do mind, as @jlsherrill says it doesn't appear to work...

Shouldn't Pulp stuff be in their own SELinux policy anyway?  I have little desire to maintain Katello's SELinux policies, let alone Pulp's.

lzap:
Sorry I mis-ready that. Will investigate.

The problem is that Satellite 6 installer configures it in a way that they dont support. I am not sure how this shell be solved nicely. It's even worse - since this port is already defined in the core policy, it shall not be removed from it's binding.

lzap:
This patch was moved to newly created `katello-selinux` repo. I've also fixed the patch there. Please move. Thanks.

PR #42, opened 2015-01-16
Fixes #8989 - Add docker_port_t port and boolean

lzap:
Boolean passenger_can_connect_docker allows connections to newly created 
docker_port_t which is not yet defined in RHEL7/Fedora. This can be used when
users starts Docker on TCP (defaults to UNIX sockets). IANA assigned 2375 and
2376 ports for http/https communication on 2015-01-09.

Denial:

type=AVC msg=audit(1421352630.245:15331): avc:  denied  { name_connect } for 
pid=4803 comm="ruby" dest=2375 scontext=unconfined_u:system_r:passenger_t:s0 
tcontext=system_u:object_r:port_t:s0 tclass=tcp_socket

@domcleal - in the patch I propose to have that boolean set to `true` by
default to help non-SELinux aware users.

### DIFF
--- a/foreman-selinux-disable
+++ b/foreman-selinux-disable
@@ -9,7 +9,7 @@ do
   if /usr/sbin/semodule -s $selinuxvariant -l >/dev/null; then
     # Remove all user defined ports (including the default one)
     /usr/sbin/semanage port -E | \
-      grep elasticsearch_port_t | \
+      grep -E '(elasticsearch|docker)_port_t' | \
       sed s/-a/-d/g | \
       /usr/sbin/semanage -S $selinuxvariant -i -
     # Unload policy
--- a/foreman-selinux-enable
+++ b/foreman-selinux-enable
@@ -22,6 +22,9 @@ do
     /usr/sbin/semanage port -E | grep -q elasticsearch_port_t || \
       echo "port -a -t elasticsearch_port_t -p tcp 9200-9300" >> $TMP
 
+    /usr/sbin/semanage port -E | grep -q docker_port_t || \
+      echo "port -a -t docker_port_t -p tcp 2375-2376" >> $TMP
+
     /usr/sbin/semanage -S $selinuxvariant -i $TMP
   fi
 done
--- a/foreman.te
+++ b/foreman.te
@@ -64,6 +64,20 @@ gen_tunable(passenger_can_connect_all, false)
 
 ## <desc>
 ## <p>
+## Determine whether passenger can connect to Docker via TCP
+## </p>
+## </desc>
+gen_tunable(passenger_can_connect_docker_tcp, true)
+
+## <desc>
+## <p>
+## Determine whether passenger can connect to Docker via local socket
+## </p>
+## </desc>
+gen_tunable(passenger_can_connect_docker_unix, true)
+
+## <desc>
+## <p>
 ## Determine whether passenger can connect to OpenStack
 ## </p>
 ## </desc>
@@ -362,6 +376,24 @@ sysnet_read_config(websockify_t)
 abrt_stream_connect(websockify_t)
 files_search_var_lib(websockify_t)
 read_files_pattern(websockify_t, puppet_var_lib_t, puppet_var_lib_t)
+
+######################################
+#
+# Docker
+#
+
+type docker_port_t;
+corenet_port(docker_port_t)
+
+tunable_policy(`passenger_can_connect_docker_tcp',`
+    allow passenger_t docker_port_t:tcp_socket name_connect;
+')
+
+tunable_policy(`passenger_can_connect_docker_unix',`
+    ifndef(`distro_rhel6', `
+        docker_stream_connect(passenger_t)
+    ')
+')
 
 ######################################
 #
--- a/Makefile
+++ b/Makefile
@@ -5,7 +5,7 @@ VARIANT=targeted
 TYPE=apps
 VERSION=99.999
 MANCHAPTER=Foreman
-TMPDIR=local-tmp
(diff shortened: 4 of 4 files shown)


lzap:
Added UNIX local sockets for RHEL7+ as well.

lzap:
@domcleal mind review & merge?

foreman-selinux-enable
```diff
@@ -22,6 +22,9 @@ do
     /usr/sbin/semanage port -E | grep -q elasticsearch_port_t || \
       echo "port -a -t elasticsearch_port_t -p tcp 9200-9300" >> $TMP
 
+    /usr/sbin/semanage port -E | grep -q docker_port_t || \
```
domcleal, line 25: also update foreman-selinux-disable

lzap, line 25: Updated, good catch.

lzap:
Hmmm this will conflict out once https://github.com/theforeman/foreman-selinux/pull/44 is merged due to the elasticport...

foreman.te
```diff
@@ -64,6 +64,20 @@ gen_tunable(passenger_can_connect_all, false)
 
 ## <desc>
 ## <p>
+## Determine whether passenger can connect to OpenStack via TCP
```
domcleal: Docker, and below

foreman-selinux-disable
```diff
@@ -9,7 +9,7 @@ do
   if /usr/sbin/semodule -s $selinuxvariant -l >/dev/null; then
     # Remove all user defined ports (including the default one)
     /usr/sbin/semanage port -E | \
-      grep elasticsearch_port_t | \
+      grep -E '(elasticsearch|docker)_port_t' | \
```
lzap, line 12: This will conflict, we need to get rid of Katello first.

foreman.te
lzap: Copy&paste error, fixed.

foreman-selinux-disable
domcleal, line 12: I'd prefer to do it this way around if I may, because I'd probably include this in a 1.7 release.  And I've already started testing... :)

And 42 < 44!

domcleal:
On IRC you're suggesting there's a problem with docker_port_t, could you please explain what it is?

foreman-selinux-disable
lzap, line 12: I don't understand you a single word here? And 42? No clue :-)

I mean we will have a _git conflict_. The extender grep will likely not be necessary.

domcleal, line 12: Sure, there will be a conflict, but you can then rebase that PR.

lzap:
Sure, I've used our internal list simply because Dan and Mirek are subscribed there. Here it is:

Foreman is supported both on RHEL6 and RHEL7. We ship our own policy
in foreman-selinux package.

For version 1.8 there are three services that Foreman application
(passenger_t) needs to connect to. These are:
- ElasticSearch
- Docker
- OpenStack (Nova and others)

In RHEL7 this is not a problem as all the ports are already defined,
except for the ElasticSearch which is not yet in Fedora.

For RHEL6 there is a problem because non of these SELinux ports are
defined yet. My initial idea which already made its way upstream is to
simply define those ports and assign port numbers to them:
- elasticsearch_port_t
- docker_port_t
- osapi_compute_port_t

When I was working on moving some rules from foreman-selinux into
katello-selinux separate package for the Katello plugin, I hit an issue.
When a port type and number are defined, it is not possible to redefine
them. And we will likely hit this problem in future.

I can think of the following situation:
- Foreman user runs an instance with our policy and all the
  ports listed above define. Everything is okay.
- Red Hat SELinux team releases an errata for Core policy that adds
  docker_port_t. Since this was already defined, the yum transaction
  will likely fail. There is no way to upgrade SELinux now on a
  Foreman instance.
- This will be quite hard to fix as port must be undefined, our policy
  must be fixed first removing the port type and new core policy must
  be upgraded then.

I've been discussing this with Mirek from SELinux team already and it
looks like this one is tough. There might be way out by providing a  
custom port (foreman_connect_port_t) that would be used to assign all
the ports into until the SELinux core policy is updated. Mirek says that 
this could be easier to upgrade, still not ideal tho.

For the record, I've asked on the SELinux Fedora list as well:

https://lists.fedoraproject.org/pipermail/selinux/2015-February/016591.html

domcleal:
Is this simply because of a possible name conflict?  IIRC if the port _numbers_ are also defined that the policies will conflict (e.g. you can't add a new port type that redefines something that's already called something else).

So even if we created a custom port type we could still conflict just on the numbers unless we always removed it immediately before the distro is updated (as if that's possible!).

lzap:
It's actually both: port type declaration and port number assignement. You cannot redefine single one.

My idea is to create our own type (ideally in some naming convention like `custom_whatever_NNNN_port_type`) so SELinux can unassign them in the upgrade scriplet. Type declaration is not a problem, they wont be declaring those again. Once core policy is upgraded, we only need to push out update for the new port type (e.g. proper `docker_port_t`). But the key thing is here we don't get any conflicts during upgrades.

domcleal:
> so SELinux can unassign them in the upgrade scriplet

Do you mean selinux-policy itself?

lzap:
Yup, that's what we've talked about on DevConf, let's see on the mailing list.

lzap:
It appears that core policy does not use semanage to define ports, they hardcode it in the policy itself. I don't want to block this, so my proposal is to merge this as is (let's define our ports) and work with them on better solution long-term.

domcleal:
Thanks @lzap!  Merged as 0d132f6.

PR #43, opened 2015-01-26
Fixes #9106 - added rule for AMQP connection

lzap:


lzap:
This conversation was moved to katello-selinux repo. Please review there.

lzap:
Since Foreman Tasks is generic plugin, I am tempted to keep it in the foreman-selinux policy. Opinions?

lzap:
Yeah let's keep it here as there are plans to use it for orchestration in the future.

lzap:
@iNecas please review here. I've heard that katello-jobs is no longer an option, should we remove it from `foreman.fc` then? Any other changes to the plugin?

domcleal:
Isn't this Katello-specific rather than foreman-tasks specific?  i.e. it's connecting to Katello/Pulp's AMPQ service from a Katello task?

iNecas:
Yes, katello is the one to add the connection to the amqp server

iNecas:
And yes, katello-jobs is no longer in use, we can remove those

lzap:
I've moved the rule into foreman-tasks section and corrected the comment. @iNecas can you confirm that those are still necessary:

```
# Foreman Tasks plugin

/usr/share/gems/gems/foreman-tasks-.*/bin/foreman-tasks -- gen_context(system_u:object_r:foreman_tasks_exec_t,s0)
/opt/rh/ruby193/root/usr/share/gems/gems/foreman-tasks-.*/bin/foreman-tasks -- gen_context(system_u:object_r:foreman_tasks_exec_t,s0)
```

Thanks

domcleal:
I think according to @iNecas' comment, that this should be in the Katello SELinux policy, not foreman-tasks.

lzap:
@iNecas please confirm, then I will move this to Katello.

iNecas:
I hereby confirm that this should go to the katello's selinux policy

lzap:
Hehe, roger wilco!

PR #44, opened 2015-01-27
Fixes #9126 - moved Katello policy to a separate repo

lzap:


lzap:
@domcleal mind merging if that's sane so we can release https://github.com/Katello/katello-selinux/pull/3

Thank you.

foreman-selinux-disable
```diff
@@ -7,11 +7,6 @@ set +e
 for selinuxvariant in targeted
 do
   if /usr/sbin/semodule -s $selinuxvariant -l >/dev/null; then
-    # Remove all user defined ports (including the default one)
```
domcleal, line 10: We need to remove this on upgrade somehow... maybe move it to the enable script?

lzap, line 10: I was thinking to leave the port as is (defined) but good point. I am not sure if SELinux will like the fact the type is now defined in a different policy. Let me test.

lzap, line 10: That's right. It will not upgrade:

```
Loading targeted modules: foreman-proxy foreman katello
libsepol.context_from_record: type elasticsearch_port_t is not defined (No such file or directory).
libsepol.context_from_record: could not create context structure (Invalid argument).
libsepol.port_from_record: could not create port structure for range 9200:9300 (tcp) (Invalid argument).
libsepol.sepol_port_modify: could not load port range 9200 - 9300 (tcp) (Invalid argument).
libsemanage.dbase_policydb_modify: could not modify record value (Invalid argument).
libsemanage.semanage_base_merge_components: could not merge local modifications into policy (Invalid argument).
/usr/sbin/semodule:  Failed!
```

Therefore we need to remove the ports on upgrade.

lzap:
So leaving it in enable script seems right, good plan. We can remove for 1.9 version (left a note there). Also fixed Makefile - now I use the remote-load for both policies and files were not being deleted.

We need to require specific version of foreman-selinux from katello-selinux. Shell I bump a minor version or something?

lzap:
I mean I could bump release to 1:

```
Version:        1.8.0
Release:        1.develop%{?dotalphatag}%{?dist}
```

lzap:
Wait, it works for the first upgrade, but then when katello module is loaded, we no longer want to remove the elasticsearch port. I am thinking of adding check for "katello" module presence (semodule list) and only remove when katello is _not_ present. What you think?

lzap:
The issue we see here is actually quite ugly - a port and number assignement cannot be redefined. I think a solution would be to introduce special port type `foreman_connect_port_t` and a rule allowing connect. Then we could assign all the "not-yet-introduced-in-rhel6" ports there. Upgrades would be easier (one just need to handle with the port assignement).

What you think of this? (See my mail to the list). This way we can solve this particular situation with `katello-selinux` refactoring.

I can make the changes there and just remove the elasticsearch_port_t here forever. No "katello" check is needed.

Although this is not that relevant for this PR, we have a Docker PR opened for this repo and that would be relevant. I would like to introduce the special port for those "workarounds for RHEL6". Opinion? Thanks.

domcleal:
I don't think the port issue particularly matters here, that's an issue for the docker PR and when you replace it in the katello-selinux PR.  It's probably best to remove it when the katello module isn't present as you suggested earlier, then this allows you to put it back in the katello module if you wish.

Typically you'd just add a dependency on >= 1.8.0 in a package, we don't ever attempt to version inside nightlies as they're in continuous development.

lzap:
Rebased, amended the change. Updated the upgrade `enable` step.

The only drawback is that `katello-selinux` will be always installed before `foreman-selinux`. Therefore during the initial upgrade, katello module will not load due the elasticsearch port issue. Then `foreman-selinux` policy will be upgraded (and port removed) and after that users need to manually reinstall the `katello-selinux` package (?)

We could do this in `%pretrans: $1 == 0` in the package as well, that would be better. Do you think this would work?

domcleal:
Interesting problem!  However I disagree that katello-selinux would always be installed first.  If katello-selinux depends on foreman-selinux, then the package manager should upgrade foreman-selinux first.

I tested the order of the scriptlets on EL7 and see that it runs pre/post for each package in turn, so this means foreman-selinux-enable will run before katello-selinux-enable (I presume), which means we might be able to remove the port declaration before the katello policy is loaded?

domcleal:
So actually, re-reading your patch, I don't see why it wouldn't work right now, assuming katello-selinux depends on foreman-selinux.

lzap:
Yeah I was thinking bad - what you said seems to be correct. Therefore no additonal snippets are necessary and this should work as is. :+1:

domcleal:
Can you link scratch builds?  I think it'd be a good idea to test.

lzap:
Rebased because of tags.

http://koji.katello.org/koji/taskinfo?taskID=230896
http://koji.katello.org/koji/taskinfo?taskID=230898

lzap:
```
[root@fsix ~]# semodule -l | grep foreman
foreman 1.8.0
[root@fsix ~]# yum upgrade 'http://koji.katello.org/koji/getfile?taskID=230899&name=foreman-selinux-1.9.0-0.develop.el6.noarch.rpm'
Loaded plugins: fastestmirror, security
Setting up Upgrade Process
getfile?taskID=230899&name=foreman-selinux-1.9.0-0.develop.el6.noarch.rpm                               |  43 kB     00:00     
Examining /var/tmp/yum-root-1TQAnN/getfile?taskID=230899&name=foreman-selinux-1.9.0-0.develop.el6.noarch.rpm: foreman-selinux-1.9.0-0.develop.el6.noarch
Marking /var/tmp/yum-root-1TQAnN/getfile?taskID=230899&name=foreman-selinux-1.9.0-0.develop.el6.noarch.rpm as an update to foreman-selinux-1.8.0-0.develop.201412151103gite2863e4.el6.noarch
Loading mirror speeds from cached hostfile
 * base: mirror.karneval.cz
 * epel: download.eng.brq.redhat.com
 * extras: mirror.karneval.cz
 * updates: mirror.karneval.cz
Resolving Dependencies
--> Running transaction check
---> Package foreman-selinux.noarch 0:1.8.0-0.develop.201412151103gite2863e4.el6 will be updated
---> Package foreman-selinux.noarch 0:1.9.0-0.develop.el6 will be an update
--> Finished Dependency Resolution

Dependencies Resolved

===============================================================================================================================
 Package          Arch    Version                Repository                                                               Size
===============================================================================================================================
Updating:
 foreman-selinux  noarch  1.9.0-0.develop.el6    /getfile?taskID=230899&name=foreman-selinux-1.9.0-0.develop.el6.noarch   79 k

Transaction Summary
===============================================================================================================================
Upgrade       1 Package(s)

Total size: 79 k
Is this ok [y/N]: y
Downloading Packages:
Running rpm_check_debug
Running Transaction Test
Transaction Test Succeeded
Running Transaction
  Updating   : foreman-selinux-1.9.0-0.develop.el6.noarch                                                                  1/2 
  Cleanup    : foreman-selinux-1.8.0-0.develop.201412151103gite2863e4.el6.noarch                                           2/2 
  Verifying  : foreman-selinux-1.9.0-0.develop.el6.noarch                                                                  1/2 
  Verifying  : foreman-selinux-1.8.0-0.develop.201412151103gite2863e4.el6.noarch                                           2/2 

Updated:
  foreman-selinux.noarch 0:1.9.0-0.develop.el6                                                                                 

Complete!
[root@fsix ~]# semodule -l | grep foreman
foreman 1.9.0
```

domcleal:
Sorry, I meant that it works with katello-selinux installation/upgrade too?

lzap:
Okay, now testing this setup:
- installed foreman and katello nightlies
- checked katello-selinux is installed and loaded
- will upgrade foreman-selinux with this patch

lzap:
```
Preparing...                ########################################### [100%]
   1:foreman-selinux        ########################################### [100%]
libsepol.context_from_record: type elasticsearch_port_t is not defined (No such file or directory).
libsepol.context_from_record: could not create context structure (Invalid argument).
libsepol.port_from_record: could not create port structure for range 9200:9300 (tcp) (Invalid argument).
libsepol.sepol_port_modify: could not load port range 9200 - 9300 (tcp) (Invalid argument).
libsemanage.dbase_policydb_modify: could not modify record value (Invalid argument).
libsemanage.semanage_base_merge_components: could not merge local modifications into policy (Invalid argument).
/usr/sbin/semanage: Could not commit semanage transaction
```

Hmmmm

lzap:
Here is my plan: in the enable script, if katello module was loaded, then unload it, remove the port, load it back.

domcleal:
I'm not entirely sure I follow, but questioning the setup you tested, does katello-selinux exist right now?  And does it define elasticsearch_port_t?

I'd expect you to either upgrade both katello-selinux and foreman-selinux in one transaction, so the port moves from one to the other, or to be upgrading foreman-selinux and installing katello-selinux for the first time.

lzap:
The problem is users will not be upgrading katello and foreman in one transaction. They will be upgrading foreman and installing (new package) for katello.

lzap:
This is the current tree (master -> nightly): https://github.com/Katello/katello-selinux

domcleal:
Wouldn't an upgrade + install be in a single yum transaction?  I'm not sure why it would be separate.

lzap:
Right let me retest with a single transaction, need to create local repo for that.

lzap:
Setup is not easy, made two mistakes, third try... :-P

domcleal:
> Right let me retest with a single transaction, need to create local repo for that.

I should also suggest "yum shell" which lets you do "install /foo.rpm", "upgrade bar" and then "run".

lzap:
NOOOOOO I am setting things up for third time now... omfg

lzap:
Does not work, after transaction the port is missing completely (the type). RHEL6:

```
[root@fsix ~]# rpm -q foreman-selinux katello-selinux
foreman-selinux-1.9.0-0.develop.201503091423git639a8ed.el6.noarch
package katello-selinux is not installed
[root@fsix ~]# semodule -l | grep -E '(foreman|katello)'
foreman 1.9.0
[root@fsix ~]# find rpmbuild/RPMS/noarch/
rpmbuild/RPMS/noarch/
rpmbuild/RPMS/noarch/foreman-proxy-selinux-1.9.1-0.develop.el6.noarch.rpm
rpmbuild/RPMS/noarch/katello-selinux-2.2.1-1.el6.noarch.rpm
rpmbuild/RPMS/noarch/foreman-selinux-1.9.1-0.develop.el6.noarch.rpm
[root@fsix ~]# yum shell
Loaded plugins: fastestmirror, security
Setting up Yum Shell
> ts

> install rpmbuild/RPMS/noarch/katello-selinux-2.2.1-1.el6.noarch.rpm
Setting up Install Process
Examining rpmbuild/RPMS/noarch/katello-selinux-2.2.1-1.el6.noarch.rpm: katello-selinux-2.2.1-1.el6.noarch
Marking rpmbuild/RPMS/noarch/katello-selinux-2.2.1-1.el6.noarch.rpm to be installed
Loading mirror speeds from cached hostfile
 * base: mirror.karneval.cz
 * epel: download.eng.brq.redhat.com
 * extras: mirror.karneval.cz
 * updates: mirror.karneval.cz
> upgrade rpmbuild/RPMS/noarch/foreman-selinux-1.9.1-0.develop.el6.noarch.rpm
Setting up Upgrade Process
Examining rpmbuild/RPMS/noarch/foreman-selinux-1.9.1-0.develop.el6.noarch.rpm: foreman-selinux-1.9.1-0.develop.el6.noarch
Marking rpmbuild/RPMS/noarch/foreman-selinux-1.9.1-0.develop.el6.noarch.rpm as an update to foreman-selinux-1.9.0-0.develop.201503091423git639a8ed.el6.noarch
> run
--> Running transaction check
---> Package foreman-selinux.noarch 0:1.9.0-0.develop.201503091423git639a8ed.el6 will be updated
---> Package foreman-selinux.noarch 0:1.9.1-0.develop.el6 will be an update
---> Package katello-selinux.noarch 0:2.2.1-1.el6 will be installed
--> Finished Dependency Resolution

===============================================================================================================================
 Package                 Arch           Version                      Repository                                           Size
===============================================================================================================================
Installing:
 katello-selinux         noarch         2.2.1-1.el6                  /katello-selinux-2.2.1-1.el6.noarch                  55 k
Updating:
 foreman-selinux         noarch         1.9.1-0.develop.el6          /foreman-selinux-1.9.1-0.develop.el6.noarch          79 k

Transaction Summary
===============================================================================================================================
Install       1 Package(s)
Upgrade       1 Package(s)

Total size: 134 k
Is this ok [y/N]: y
Downloading Packages:
Running rpm_check_debug
Running Transaction Test
Transaction Test Succeeded
Running Transaction
  Updating   : foreman-selinux-1.9.1-0.develop.el6.noarch                                                                  1/3 
  Installing : katello-selinux-2.2.1-1.el6.noarch                                                                          2/3 
Non-fatal POSTIN scriptlet failure in rpm package katello-selinux-2.2.1-1.el6.noarch
/usr/sbin/semanage: Type elasticsearch_port_t is invalid, must be a port type
warning: %post(katello-selinux-2.2.1-1.el6.noarch) scriptlet failed, exit status 1
  Cleanup    : foreman-selinux-1.9.0-0.develop.201503091423git639a8ed.el6.noarch                                           3/3 
  Verifying  : katello-selinux-2.2.1-1.el6.noarch                                                                          1/3 
  Verifying  : foreman-selinux-1.9.1-0.develop.el6.noarch                                                                  2/3 
  Verifying  : foreman-selinux-1.9.0-0.develop.201503091423git639a8ed.el6.noarch                                           3/3 

Installed:
  katello-selinux.noarch 0:2.2.1-1.el6                                                                                         

Updated:
  foreman-selinux.noarch 0:1.9.1-0.develop.el6                                                                                 

Finished Transaction
> Leaving Shell
[root@fsix ~]# semanage boolean -l | grep elastic
[root@fsix ~]# katello-selinux-enable 
/usr/sbin/semanage: Type elasticsearch_port_t is invalid, must be a port type
[root@fsix ~]# sesearch --allow -s elasticsearch_port_t -C
ERROR: could not find datum for type elasticsearch_port_t
```

Very same result on RHEL7.

domcleal:
Does the katello policy you're testing redefine the elasticsearch_port_t type?  It isn't in katello-selinux's master branch, but I don't know if you're testing locally with some other changes.

https://github.com/theforeman/foreman-selinux/blob/1.8.0-RC1/foreman.te#L405 is the definition in our policy which you're correctly removing in this PR, but it will then need redefining in katello.te instead.

lzap:
Yeah I am testing that currently, it looks like we need a change in katello-selinux, hold on.

lzap:
Ok tested but one change is needed on the katello side. I haven't realized I actually do not create the type anymore there: https://github.com/theforeman/foreman-selinux/pull/44

This is test from EL7, the very same result on EL6:

```
[root@fseven ~]# semodule -l | grep -E '(foreman|katello)'
foreman 1.8.0
[root@fseven ~]# yum shell
Loaded plugins: fastestmirror
> ts

> install katello-selinux-2.2.1-1.git.0.118c360.el7.noarch.rpm
Examining katello-selinux-2.2.1-1.git.0.118c360.el7.noarch.rpm: katello-selinux-2.2.1-1.git.0.118c360.el7.noarch
Marking katello-selinux-2.2.1-1.git.0.118c360.el7.noarch.rpm to be installed
> upgrade rpmbuild/RPMS/noarch/foreman-selinux-1.9.1-0.develop.el7.centos.noarch.rpm
Loading mirror speeds from cached hostfile
 * base: ftp.cvut.cz
 * epel: mirror.hosting90.cz
 * extras: ftp.cvut.cz
 * updates: ftp.cvut.cz
Examining rpmbuild/RPMS/noarch/foreman-selinux-1.9.1-0.develop.el7.centos.noarch.rpm: foreman-selinux-1.9.1-0.develop.el7.centos.noarch
Marking rpmbuild/RPMS/noarch/foreman-selinux-1.9.1-0.develop.el7.centos.noarch.rpm as an update to foreman-selinux-1.8.0-0.1.RC1.el7.noarch
> run
--> Running transaction check
---> Package foreman-selinux.noarch 0:1.8.0-0.1.RC1.el7 will be updated
---> Package foreman-selinux.noarch 0:1.9.1-0.develop.el7.centos will be an update
---> Package katello-selinux.noarch 0:2.2.1-1.git.0.118c360.el7 will be installed
--> Finished Dependency Resolution

===============================================================================================================================
 Package             Arch       Version                           Repository                                              Size
===============================================================================================================================
Installing:
 katello-selinux     noarch     2.2.1-1.git.0.118c360.el7         /katello-selinux-2.2.1-1.git.0.118c360.el7.noarch       55 k
Updating:
 foreman-selinux     noarch     1.9.1-0.develop.el7.centos        /foreman-selinux-1.9.1-0.develop.el7.centos.noarch      77 k

Transaction Summary
===============================================================================================================================
Install  1 Package
Upgrade  1 Package

Total size: 132 k
Is this ok [y/d/N]: y
Downloading packages:
Running transaction check
Running transaction test
Transaction test succeeded
Running transaction
Warning: RPMDB altered outside of yum.
  Updating   : foreman-selinux-1.9.1-0.develop.el7.centos.noarch                                                           1/3 
  Installing : katello-selinux-2.2.1-1.git.0.118c360.el7.noarch                                                            2/3 
  Cleanup    : foreman-selinux-1.8.0-0.1.RC1.el7.noarch                                                                    3/3 
  Verifying  : katello-selinux-2.2.1-1.git.0.118c360.el7.noarch                                                            1/3 
  Verifying  : foreman-selinux-1.9.1-0.develop.el7.centos.noarch                                                           2/3 
  Verifying  : foreman-selinux-1.8.0-0.1.RC1.el7.noarch                                                                    3/3 

Installed:
  katello-selinux.noarch 0:2.2.1-1.git.0.118c360.el7                                                                           

Updated:
  foreman-selinux.noarch 0:1.9.1-0.develop.el7.centos                                                                          

Finished Transaction
> Leaving Shell
[root@fseven ~]# semodule -l | grep -E '(foreman|katello)'
foreman 1.9.1
katello 2.2.1
```

foreman-selinux-enable
```diff
@@ -13,15 +13,22 @@ trap "rm -rf '$TMP'" EXIT INT TERM
 for selinuxvariant in targeted
 do
   if /usr/sbin/semodule -s $selinuxvariant -l >/dev/null; then
+    # UPGRADE STEP for Foreman 1.8 (remove after release)
+    # If katello policy was not yet loaded and elasticsearch port is still defined,
+    # remove it (it was moved to katello policy). Katello policy will not load before
+    # we remove the port definition (because SELinux prevents that).
+    semodule -l | grep -q katello || \
```
domcleal: Is katello-selinux already released?  It sounds like it has been, which could be an issue.

If the existing katello module is currently loaded, then this won't remove elasticsearch_port_t.  You might need to handle it a bit more carefully, as if the package has been released, existing versions won't have added the port and you still need to remove it.

lzap: It will work because released katello selinux (2.2) does not define any type. Only the fix which was merged recently introduces it.

Several nightly installation upgrades might be broken since the patch already made it to katello nightlies. I was hoping to shorten this window by merging this as well. Haven't checked github messages for few days now.

lzap: Oh the patch https://github.com/Katello/katello-selinux/issues/6 was not yet merged, that's even better. Current katello-selinux implementation is harmless, it does not define the type, only requires it.

domcleal: Requiring it is a problem though.  As the current released katello-selinux requires it without defining it, then this will happen I think:
1. foreman-selinux-enable runs
   1. does not remove elasticsearch_port_t as 'katello' module is loaded
   2. loads new 'foreman' module which no longer defines it, this will throw an error as the 'katello' policy references it

While what should have happened, if the current released katello-selinux wasn't using it or didn't exist (the scenario you tested) is:
1. foreman-selinux-enable runs
   1. removes elasticsearch port range as 'katello' isn't loaded
   2. loads new 'foreman' module which no longer defines it
2. katello-selinux-enable runs
   1. loads new 'katello' module which defines it
   2. adds elasticsearch_port_t range

lzap: The katello has all elasticsearch things in optional block, let me retest with the scenario you just described. I think it should work.

domcleal: Ah, that might work then.  Let me know how it goes!

lzap: It does not:

```
[root@fsix ~]# rpm -Uvh --force https://fedorapeople.org/groups/katello/releases/yum/nightly/katello/RHEL/6Server/x86_64/katello-selinux-2.2.1-1.el6.noarch.rpm
Retrieving https://fedorapeople.org/groups/katello/releases/yum/nightly/katello/RHEL/6Server/x86_64/katello-selinux-2.2.1-1.el6.noarch.rpm
Preparing...                ########################################### [100%]
   1:katello-selinux        ########################################### [100%]
[root@fsix ~]# rpm -Uvh --force "http://koji.katello.org/koji/getfile?taskID=236410&name=foreman-selinux-1.9.0-0.develop.el6.noarch.rpm"
Retrieving http://koji.katello.org/koji/getfile?taskID=236410&name=foreman-selinux-1.9.0-0.develop.el6.noarch.rpm
Preparing...                ########################################### [100%]
   1:foreman-selinux        ########################################### [100%]
libsepol.context_from_record: type elasticsearch_port_t is not defined (No such file or directory).
libsepol.context_from_record: could not create context structure (Invalid argument).
libsepol.port_from_record: could not create port structure for range 9200:9300 (tcp) (Invalid argument).
libsepol.sepol_port_modify: could not load port range 9200 - 9300 (tcp) (Invalid argument).
libsemanage.dbase_policydb_modify: could not modify record value (Invalid argument).
libsemanage.semanage_base_merge_components: could not merge local modifications into policy (Invalid argument).
/usr/sbin/semanage: Could not commit semanage transaction
```

lzap: Ok is the right solution to remove katello module during the upgrade?
- foreman-selinux-enable runs
  - remove katello module (katello-selinux-disable)
  - removes elasticsearch port range
  - loads new 'foreman' module which no longer defines it
- katello-selinux-enable runs
  - loads new 'katello' module which defines it
  - adds elasticsearch_port_t range

lzap: It works:

```
[root@fsix ~]# /usr/sbin/katello-selinux-disable
[root@fsix ~]# rpm -Uvh --force "http://koji.katello.org/koji/getfile?taskID=236410&name=foreman-selinux-1.9.0-0.develop.el6.noarch.rpm"
Retrieving http://koji.katello.org/koji/getfile?taskID=236410&name=foreman-selinux-1.9.0-0.develop.el6.noarch.rpm
Preparing...                ########################################### [100%]
   1:foreman-selinux        ########################################### [100%]
```

domcleal: It might work, but we'd have to enable it after too?  (as katello-selinux may not be upgraded)

domcleal: What about not using elasticsearch_port_t in katello-selinux, so we remove all traces of it here, and then that package uses an entirely different name, like katello_allowed_ports_t.

lzap:
Looks like elasticsearch was removed from katello nightly now, so we can proceed with this patch now. I will rebase it.

lzap:
Rebased, things are much more simple. I added a block both here and in Katello to remove user-defined elasticsearch_port_t types, because we no longer need it. We need to have at least one release with this upgrade command, then we can get rid of it.

foreman-selinux-enable
```diff
@@ -13,15 +13,20 @@ trap "rm -rf '$TMP'" EXIT INT TERM
 for selinuxvariant in targeted
 do
   if /usr/sbin/semodule -s $selinuxvariant -l >/dev/null; then
+
+    # Remove previously defined elasticsearch_port_t
+    # (this can be removed in future release)
+    /usr/sbin/semanage port -E | \
+      grep elasticsearch_port_t | \
+      sed s/-a/-d/g | \
+      /usr/sbin/semanage -S $selinuxvariant -i -
```
domcleal: Nitpick, if you can only run `semanage -i` here if the port exists then it'll save a lot of installation time (on every single patch release for every user) as these calls are expensive.  Probably changing the mktemp call to add `-d`, then save the `port -E` output to only call it when required.

You could even re-use the port -E output below, which should keep the speed the same.

lzap: Sure, I directly added that to the semanage temp file which should work just fine.

lzap:
Amended, if you can do me a favor and only merge this _after_ the openstack PR https://github.com/theforeman/foreman-selinux/pull/45 since the other one is priority for me, there will be conflicts. Thanks!

foreman-selinux-enable
```diff
@@ -13,15 +14,19 @@ trap "rm -rf '$TMP'" EXIT INT TERM
...
     # Load policy
     /usr/sbin/semanage module -S $selinuxvariant \
       -a /usr/share/selinux/${selinuxvariant}/foreman.pp.bz2
 
     echo "boolean -m --on httpd_setrlimit" > $TMP
 
-    /usr/sbin/semanage port -E | grep -q elasticsearch_port_t || \
-      echo "port -a -t elasticsearch_port_t -p tcp 9200-9300" >> $TMP
-
     /usr/sbin/semanage port -E | grep -q docker_port_t || \
```
domcleal: Use $TMP_PORTS here too.

lzap: Note to myself - I will add TMP_PORTS for all statements, including the one that will appear after I rebase this on top of `develop`.

domcleal:
Could you rebase this PR please?

lzap:
Sure thing. Done!

foreman-selinux-enable
```diff
@@ -17,20 +18,22 @@ is_redhat_6() {
...
       -a /usr/share/selinux/${selinuxvariant}/foreman.pp.bz2
 
     echo "boolean -m --on httpd_setrlimit" > $TMP
 
-    /usr/sbin/semanage port -E | grep -q elasticsearch_port_t || \
-      echo "port -a -t elasticsearch_port_t -p tcp 9200-9300" >> $TMP
-
-    /usr/sbin/semanage port -E | grep -q docker_port_t || \
-      echo "port -a -t docker_port_t -p tcp 2375-2376" >> $TMP
+    grep docker_port_t $TMP_PORTS | echo "port -a -t docker_port_t -p tcp 2375-2376" >> $TMP
```
domcleal: Should `|` be `||` here, so as to create the port only when `grep` returns a bad error code?  I don't think the output of grep needs piping.  Also prefer `grep -q`.

foreman-selinux-enable
```diff
@@ -17,20 +18,22 @@ is_redhat_6() {
 for selinuxvariant in targeted
 do
   if /usr/sbin/semodule -s $selinuxvariant -l >/dev/null; then
+    /usr/sbin/semanage port -E > $TMP_PORTS
+
+    # Remove previously defined elasticsearch_port_t
+    # (this can be removed in future release)
+    grep elasticsearch_port_t $TMP_PORTS | sed s/-a/-d/g >> $TMP
```
domcleal: I think this needs to be executed _before_ the policy is loaded, otherwise the port can't be removed:

<pre>
+ for selinuxvariant in targeted
+ /usr/sbin/semodule -s targeted -l
+ /usr/sbin/semanage port -E
+ sed s/-a/-d/g
+ grep elasticsearch_port_t /tmp/foreman-selinux-enable.r7p4qWmmp9
+ /usr/sbin/semanage module -S targeted -a /usr/share/selinux/targeted/foreman.pp.bz2
libsepol.context_from_record: type elasticsearch_port_t is not defined (No such file or directory).
libsepol.context_from_record: could not create context structure (Invalid argument).
libsepol.port_from_record: could not create port structure for range 9200:9300 (tcp) (Invalid argument).
libsepol.sepol_port_modify: could not load port range 9200 - 9300 (tcp) (Invalid argument).
libsemanage.dbase_policydb_modify: could not modify record value (Invalid argument).
libsemanage.semanage_base_merge_components: could not merge local modifications into policy (Invalid argument).
/usr/sbin/semanage: Could not commit semanage transaction
</pre>

foreman-selinux-enable
lzap: Indeed, that was a typo.

foreman-selinux-enable
lzap: Oh sure, we need to remember to remove it, it's one extra (slow) call.

domcleal: But `semanage ... -d` can be called only when the port exists in $TMP_PORTS (which we need to call anyway for docker_port_t etc).  If it doesn't exist, we can avoid a performance penalty.

lzap: How about this - there are now two temp files and they only gets executed when non-empty. I haven't tested this but in case there is no typo this should work and be flexible for the future.

lzap: > But `semanage ... -d` can be called only when the port exists in $TMP_PORTS (which we need to call anyway for docker_port_t etc).  If it doesn't exist, we can avoid a performance penalty.

Right, my latest push should do it.

## 

Later,
 Lukas #lzap Zapletal

foreman-selinux-enable
```diff
@@ -17,23 +19,24 @@ is_redhat_6() {
 for selinuxvariant in targeted
 do
   if /usr/sbin/semodule -s $selinuxvariant -l >/dev/null; then
-    # Load policy
-    /usr/sbin/semanage module -S $selinuxvariant \
-      -a /usr/share/selinux/${selinuxvariant}/foreman.pp.bz2
+    /usr/sbin/semanage port -E > $TMP_PORTS
```
domcleal, line 22: The redirection here doesn't seem to work on EL6 with policycoreutils-python-2.0.83-24.el6.

<pre>
# rpm -Uvh foreman-selinux-1.12.0-0.develop.201605040716gitb1daca47.el6.noarch.rpm
Preparing...                ########################################### [100%]
   1:foreman-selinux        ########################################### [100%]
libsepol.context_from_record: type elasticsearch_port_t is not defined (No such file or directory).
libsepol.context_from_record: could not create context structure (Invalid argument).
libsepol.port_from_record: could not create port structure for range 9200:9300 (tcp) (Invalid argument).
libsepol.sepol_port_modify: could not load port range 9200 - 9300 (tcp) (Invalid argument).
libsemanage.dbase_policydb_modify: could not modify record value (Invalid argument).
libsemanage.semanage_base_merge_components: could not merge local modifications into policy (Invalid argument).
/usr/sbin/semanage: Could not commit semanage transaction
/usr/sbin/semanage: Port tcp/2375-2376 already defined
warning: %post(foreman-selinux-1.12.0-0.develop.201605040716gitb1daca47.el6.noarch) scriptlet failed, exit status 1
[ edited it and added a cat, uncommented the trap ]
# bash -x /usr/sbin/foreman-selinux-enable
+ set +e
++ mktemp -t foreman-selinux-enable.XXXXX
+ TMP_EXEC_BEFORE=/tmp/foreman-selinux-enable.56Ryo
++ mktemp -t foreman-selinux-enable.XXXXX
+ TMP_EXEC_AFTER=/tmp/foreman-selinux-enable.BhL8L
++ mktemp -t foreman-selinux-enable.XXXXX
+ TMP_PORTS=/tmp/foreman-selinux-enable.9VQCl
+ for selinuxvariant in targeted
+ /usr/sbin/semodule -s targeted -l
+ /usr/sbin/semanage port -E
+ cat /tmp/foreman-selinux-enable.9VQCl
+ sed s/-a/-d/g
+ grep elasticsearch_port_t /tmp/foreman-selinux-enable.9VQCl
+ echo 'boolean -m --on httpd_setrlimit'
+ grep -q docker_port_t /tmp/foreman-selinux-enable.9VQCl
+ echo 'port -a -t docker_port_t -p tcp 2375-2376'
+ is_redhat_6
++ rpm -q --whatprovides redhat-release --qf '%{version}'
+ test x6 = x6
+ grep -q foreman_osapi_compute_port_t /tmp/foreman-selinux-enable.9VQCl
+ echo 'port -a -t foreman_osapi_compute_port_t -p tcp 8774'
+ test -s /tmp/foreman-selinux-enable.56Ryo
+ /usr/sbin/semanage module -S targeted -a /usr/share/selinux/targeted/foreman.pp.bz2
libsepol.context_from_record: type elasticsearch_port_t is not defined (No such file or directory).
libsepol.context_from_record: could not create context structure (Invalid argument).
libsepol.port_from_record: could not create port structure for range 9200:9300 (tcp) (Invalid argument).
libsepol.sepol_port_modify: could not load port range 9200 - 9300 (tcp) (Invalid argument).
libsemanage.dbase_policydb_modify: could not modify record value (Invalid argument).
libsemanage.semanage_base_merge_components: could not merge local modifications into policy (Invalid argument).
/usr/sbin/semanage: Could not commit semanage transaction
+ test -s /tmp/foreman-selinux-enable.BhL8L
+ /usr/sbin/semanage -S targeted -i /tmp/foreman-selinux-enable.BhL8L
/usr/sbin/semanage: Port tcp/2375-2376 already defined
# ll /tmp/foreman-selinux-enable.9VQCl
-rw-------. 1 root root 0 May  4 08:44 /tmp/foreman-selinux-enable.9VQCl
# semanage port -E
port -a -t docker_port_t -p tcp 2375-2376
port -a -t elasticsearch_port_t -p tcp 9200-9300
# semanage port -E > /tmp/a
# ls -l /tmp/a
-rw-r--r--. 1 root root 0 May  4 08:46 /tmp/a
# semanage port -E | tee /tmp/a
port -a -t docker_port_t -p tcp 2375-2376
port -a -t elasticsearch_port_t -p tcp 9200-9300
# ls -l /tmp/a
-rw-r--r--. 1 root root 91 May  4 08:47 /tmp/a
</pre>

domcleal, line 22: I think the issue is simply SELinux with fd redirection - semanage runs in a context where it can't write to user_tmp_t.  Switching to permissive allows it.

lzap, line 22: I have just built on CentOS 6 (latest stable with updates) and I don't see any errors there. This was a clean install, therefore clean upgrade of the RPM package that I've built. No errors at all. `policycoreutils-python-2.0.83-24.el6.x86_64`

domcleal, line 22: Thanks for checking, it seems to have been added to the base policy between selinux-policy-targeted-3.7.19-195.el6 (EL6.4) and selinux-policy-targeted-3.7.19-231.el6 (EL6.5), so it's just my out of date installation.  That's fine, we should just bump the min version in the RPM after this is merged.

domcleal:
Merged as 96804f3, thanks @lzap.

PR #45, opened 2015-02-05
Fixes #10443 - added nova/glance/neutron rules

lzap:
Ok I've missed several other ports. The discussion is in the linked BZ.

There is one snag - the commented macro is missing in RHEL6 and we need to,
once again, define our own port. It will likely cause issues because I expect
this one to be introduced in RHEL6. I need to find a way how to deal with this
first. Then let's get back to this.

### DIFF
--- a/foreman-selinux-enable
+++ b/foreman-selinux-enable
@@ -4,6 +4,10 @@ set +e
 TMP=$(mktemp -t foreman-selinux-enable.XXXXXXXXXX)
 trap "rm -rf '$TMP'" EXIT INT TERM
 
+is_redhat_6() {
+  test x$(rpm -q --whatprovides redhat-release --qf '%{version}') = x6
+}
+
 # Load or upgrade foreman policy and set booleans.
 #
 # Dependant booleans must be managed in a separate transaction.
@@ -24,6 +28,11 @@ do
 
     /usr/sbin/semanage port -E | grep -q docker_port_t || \
       echo "port -a -t docker_port_t -p tcp 2375-2376" >> $TMP
+
+    if is_redhat_6; then
+      /usr/sbin/semanage port -E | grep -q foreman_osapi_compute_port_t || \
+        echo "port -a -t foreman_osapi_compute_port_t -p tcp 8774" >> $TMP
+    fi
 
     /usr/sbin/semanage -S $selinuxvariant -i $TMP
   fi
--- a/foreman.te
+++ b/foreman.te
@@ -142,6 +142,9 @@ files_pid_file(foreman_var_run_t)
 type foreman_proxy_port_t;
 corenet_port(foreman_proxy_port_t)
 
+type foreman_osapi_compute_port_t;
+corenet_port(foreman_osapi_compute_port_t)
+
 require{
     type bin_t;
     type httpd_t;
@@ -308,9 +311,15 @@ optional_policy(`
 
 tunable_policy(`passenger_can_connect_openstack',`
     ifdef(`distro_rhel6', `
+        # keystone (identity service)
         corenet_tcp_connect_commplex_port(passenger_t)
+        # all other ports not yet defined on rhel6
+        allow passenger_t foreman_osapi_compute_port_t:tcp_socket name_connect;
     ',`
+        # keystone (identity service)
         corenet_tcp_connect_commplex_main_port(passenger_t)
+        # nova (compute service)
+        corenet_tcp_connect_osapi_compute_port(passenger_t)
     ')
 ')
 
--- a/foreman-selinux-disable
+++ b/foreman-selinux-disable
@@ -9,7 +9,7 @@ do
   if /usr/sbin/semodule -s $selinuxvariant -l >/dev/null; then
     # Remove all user defined ports (including the default one)
     /usr/sbin/semanage port -E | \
-      grep -E '(elasticsearch|docker)_port_t' | \
+      grep -E '(elasticsearch|docker|foreman_osapi_compute)_port_t' | \
       sed s/-a/-d/g | \
       /usr/sbin/semanage -S $selinuxvariant -i -
     # Unload policy


domcleal:
Administrative note, please don't reference a closed + shipped ticket - create a new one and describe the issue.

lzap:
> Administrative note, please don't reference a closed + shipped ticket - create a new one and describe the issue.

Not so sure about this one. The original patch did not actually fix the
issue. The linked BZ failed verification and returned on me. I don't
want to introduce confusion between RM-BZ, thus I've decided to simply
ref the original one.

## 

Later,
 Lukas #lzap Zapletal

domcleal:
The Redmine issue has shipped, please file a new issue for it.

lzap:
This patch introduces new foreman_allowed_port_t type for all 
foreman-defined ports. Also it introduces helper function in the enable 
script which is now the most sane way of adding ports - if a port number is
already defined, it uses `-m` option to redefine it.

Fixed issue reference.

foreman-selinux-enable
```diff
@@ -4,6 +4,21 @@ set +e
...
+    if /usr/sbin/semanage port -E | grep -qEe "-p (tcp|udp) $2"; then
+      echo "port -m -t $1 -p tcp $2"
+    else
+      echo "port -a -t $1 -p tcp $2"
+    fi
+  fi
+}
+
+is_redhat_6() {
+  uname -r | grep -o el6 &>/dev/null
```
lzap: Are you aware of better way identifying RHEL version?

domcleal: `test x$(rpm -q --whatprovides redhat-release --qf '%{version}') = x6` would be more standard I believe

foreman.te
```diff
@@ -293,10 +299,20 @@ optional_policy(`
 
 tunable_policy(`passenger_can_connect_openstack',`
     ifdef(`distro_rhel6', `
+        # keystone (identity service)
         corenet_tcp_connect_commplex_port(passenger_t)
+        # nova (compute service)
+        # via foreman_allowed_port_t
```
domcleal: The only problem with this is that the boolean won't have an effect on EL6, while it works for EL7 and all other services.  It should probably be separate?

domcleal:
Have you checked this builds on F19?

foreman-selinux-enable
lzap: Fixed.

foreman.te
lzap: What you mean? It works both on 6 and 7.

lzap:
Not checked the build, checking. Amended the rhel6 test function.

foreman.te
domcleal: If you disable `passenger_can_connect_openstack` on EL6, it can still connect to Nova.

lzap:
Ok it looks like I tested against newer versions of RHEL6, in koji it fails. The same for F19.

I am going to change the strategy. Instead of quering for RHEL6 version in the enable script, I will query per port. If a port is not assigned (e.g. neutron) I will assign to `foreman_allowed_port_t`. Opinions?

lzap:
In the policy, I'd do something like:

```
# port is assigned via foreman_allowed_port_t if this fails
optional_policy(`
    tunable_policy(`passenger_can_connect_openstack',`
        # keystone (identity service)
        corenet_tcp_connect_commplex_port(passenger_t)
    ')
')
```

for all ports.

lzap:
Pushed the idea, _untested_. What you think?

domcleal:
Not really sure where you're going with this - so there's now a boolean, but the port might also be in foreman_allowed_port_t?  So to disable it you have to edit the port list and disable the boolean, as doing only one might just cause it to be allowed through the second..

lzap:
Ok I get your point - boolean must work on all platform. For this reason, I will introduce `foreman_openstack_port_t` and put this in the `tunable_policy` block.

Now, the solution with the `optional` blocks does not work. I haven't realized that syntax errors are not caught by this macro. Scratch that.

We have a problem here - we build on RHEL 6.3 but some OpenStack bits were added in later versions (6.6). For this reason, I think I need to do a fallback solution - everything that is not defined in 6.3 we will re-assign into our own port type.

lzap:
Something like this (untested, does not yet build).

lzap:
FYI The code is just an idea, still building and finding what is defined and what not.

lzap:
Ok I have amended and simplified everything. It turns out we only need two ports to let Foreman to work properly: 5000 and 8774 (http://docs.openstack.org/kilo/config-reference/content/firewalls-default-ports.html).

Therefore I simplified everything and doublechecked what is defined in RHEL 6.3+ and in RHEL 7.0+. This is the result.

I am going to test builds from Koji tomorrow.

lzap:
Ok this is ready for review. Works fine:

http://koji.katello.org/koji/taskinfo?taskID=300056
http://koji.katello.org/koji/taskinfo?taskID=300071
http://koji.katello.org/koji/taskinfo?taskID=300086

ohadlevy:
@lzap ping ?

lzap:
Ok I have renamed the port type name to `foreman_osapi_compute_port_t` as per our IRC discussion. This confirms to naming convention.

foreman-selinux-enable
```diff
@@ -4,6 +4,21 @@ set +e
 TMP=$(mktemp -t foreman-selinux-enable.XXXXXXXXXX)
 trap "rm -rf '$TMP'" EXIT INT TERM
 
+# Assign port number or change existing port definition.
+assign_or_change_existing() {
+  if ! /usr/sbin/semanage port -E | grep -qEe "${1}.*-p (tcp|udp) ${2}"; then
+    if /usr/sbin/semanage port -E | grep -qEe "-p (tcp|udp) $2"; then
+      echo "port -m -t $1 -p tcp $2"
```
domcleal: Is this ever needed?  I don't see how it'd be called - might be worth losing it if we can, since semanage isn't particularly fast and we can save a few seconds of installation time.

lzap: Resurrecting this one after some time, sorry for the delay. What you mean by that? Semanage is indeed slow as hell, but what can we do about it? We need to assign or change existing port, otherwise the policy won't be effective.

lzap:
I have rebased this patch, please elaborate on that comment I don't understand what your concerns are.

foreman-selinux-enable
domcleal: I think I was asking, why have this code to run semanage and reassign ports?  The two existing ports, elasticsearch and docker are both added every time around line 40, which works fine.  It looks like there's no need to check if the port exists already.

lzap: So the reasoning behind this is I expect this port to be introduced in SELinux core policy for RHEL6. Then this will cause a problem as you can't add an existing port number or range - you need explicitly to provide `-m` flag. This was coded as improvement for the future, I could perhaps extend this to the other two types. You have right point this makes things slower - typical semanage port call is about 1-2 seconds. I can perhaps store the list of ports once in a temp file and then grep it multiple times.

On the other hand, we are getting rid both of docker and elasticsearch quite soon. I can perhaps just file a ticket to make this faster. Or I can completely drop this helper and we can solve this as we hit it (if we hit it).

lzap: To be fair, I need to say that the function is not ideal, it has problems with port ranges - wont work for overlapping ranges.

domcleal: > So the reasoning behind this is I expect this port to be introduced in SELinux core policy for RHEL6. Then this will cause a problem as you can't add an existing port number or range - you need explicitly to provide -m flag

If the port's added to the base policy then I don't think this will work, because `semanage -E` is only going to list ports that customisable (i.e. the ones we add, not base policy ones). It's always going to be empty, and always go to the `port -a` case.

I do vaguely remember there being the possibility of a conflict with base policy, but I can't reproduce it on EL7 at the moment - it seems to let me redefine a port in base policy.  Possibly the reverse isn't true, that if there's a port and base policy _adds_ the same, it fails - but I don't have evidence for it.

> Or I can completely drop this helper and we can solve this as we hit it (if we hit it).

I'd suggest this for now.

lzap:
Removed the helper, now it should be fast.

foreman-selinux-enable
```diff
@@ -25,6 +29,10 @@ do
     /usr/sbin/semanage port -E | grep -q docker_port_t || \
       echo "port -a -t docker_port_t -p tcp 2375-2376" >> $TMP
 
+    if is_redhat_6; then
+      echo "port -a -t foreman_osapi_compute_port_t -p tcp 8774" >> $TMP
```
domcleal: The same `semanage port -E` check is needed here to skip the step if it's already defined (on EL6), else running this script multiple times will result in errors:

<pre>
# /usr/sbin/foreman-selinux-enable
/usr/sbin/semanage: Port tcp/8774 already defined
# echo $?
1
</pre>


The code I wanted removed was only the modification stuff.

lzap: Ok added, note I am not adding TMP_PORTS improvement in this patch, I will rebase the other one and add it here once we merge this one.

domcleal:
Merged as e54934d, thanks @lzap.

PR #47, opened 2015-03-19
Fixes #9825 - foreman can connect to remote SMTP

lzap:
I enabled this by default.

### DIFF
--- a/foreman.te
+++ b/foreman.te
@@ -97,6 +97,13 @@ gen_tunable(passenger_can_spawn_ssh, true)
 ## </desc>
 gen_tunable(passenger_can_connect_libvirt, true)
 
+## <desc>
+## <p>
+## Determine whether passenger can connect to remote SMTP services
+## </p>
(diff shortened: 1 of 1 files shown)


foreman.te
```diff
@@ -283,6 +290,10 @@ optional_policy(`
     mta_send_mail(passenger_t)
 ')
 
+tunable_policy(`passenger_can_connect_smtp',`
+    corenet_tcp_connect_smtp_port(passenger_t)
```
stbenjam, line 294: Does this cover 465 and 587 too?

domcleal, line 294: It appears so.

<pre>
# semanage port -l | grep smtp_port_t
smtp_port_t                    tcp      25, 465, 587
</pre>

lzap, line 294: Yup, sorry for the delay.

domcleal:
Merged as 2112b96, thanks @lzap.

PR #49, opened 2015-08-12
refs #4841 - change ruby193 SCL paths to tfm SCL

domcleal:
Dependent on https://github.com/theforeman/foreman-packaging/pull/249.

### DIFF
--- a/foreman.fc
+++ b/foreman.fc
@@ -39,8 +39,8 @@
 
 # Passenger SCL file contexts
 
-/opt/rh/ruby193/root/usr/(share/gems/gems|lib64/gems/exts)/passenger-.*/agents/.* -- gen_context(system_u:object_r:passenger_exec_t,s0)
-/opt/rh/ruby193/root/usr/(share/gems/gems|lib64/gems/exts)/passenger-.*/helper-scripts/prespawn -- gen_context(system_u:object_r:passenger_exec_t,s0)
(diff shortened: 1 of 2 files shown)


foreman.fc
```diff
@@ -39,8 +39,8 @@
 
 # Passenger SCL file contexts
 
-/opt/rh/ruby193/root/usr/(share/gems/gems|lib64/gems/exts)/passenger-.*/agents/.* -- gen_context(system_u:object_r:passenger_exec_t,s0)
-/opt/rh/ruby193/root/usr/(share/gems/gems|lib64/gems/exts)/passenger-.*/helper-scripts/prespawn -- gen_context(system_u:object_r:passenger_exec_t,s0)
+/opt/theforeman/tfm/root/usr/(share/gems/gems|lib64/gems/exts)/passenger-.*/agents/.* -- gen_context(system_u:object_r:passenger_exec_t,s0)
+/opt/theforeman/tfm/root/usr/(share/gems/gems|lib64/gems/exts)/passenger-.*/helper-scripts/prespawn -- gen_context(system_u:object_r:passenger_exec_t,s0)
```
lzap, line 43: Don't we use passenger from RHSCL?

Looks good, ping when this needs to be merged.

domcleal, line 43: No, we've always used our own build of it and will continue to for now at least.  There is a passenger40 collection, but adopting that's a whole other set of changes!

domcleal, line 43: > Looks good, ping when this needs to be merged.

Thanks, this may be merged now as Foreman core is built under 'tfm'.

lzap:
Merged as 3a6366ab5520afefe796ec9d1c7f99d0b23273d0, thank you!

PR #50, opened 2015-08-26
Fixes/refs #11572 - defaut spool directory has wrong context

rodehoed:
Added openscap as tunable module for foreman-proxy


domcleal:
Closing this, as I think your second PR, #51 replaces it.

PR #53, opened 2015-09-23
Fixes #11934 - docker rules are in optional block now

lzap:


domcleal:
If you install the Docker SELinux policy after, does this optional block get included again?  Or do you need to reload Foreman's policy?

lzap:
They have pulled Docker out the base policy, it is now a separate package. Once it is loaded, the whole policy is reloaded, thus our block will be included. We will likely need to require: docker-selinux in the package SPEC.

http://koji.katello.org/koji/taskinfo?taskID=348626

domcleal:
Yes, that's what I was wondering and hoping for, thanks.  We'd probably just make the foreman_docker RPM depend on it.

lzap:
Ok its a subpackage of docker, now I understand. The moment docker policy is loaded then our rule gets added.

lzap:
Tested.

```
[root@smqa-r210-06-vm04 ~]# rpm -Uvh --force "http://koji.katello.org/koji/getfile?taskID=348627&name=foreman-selinux-1.10.0-0.develop.el7.noarch.rpm"
Retrieving http://koji.katello.org/koji/getfile?taskID=348627&name=foreman-selinux-1.10.0-0.develop.el7.noarch.rpm
Preparing...                          ################################# [100%]
Updating / installing...
   1:foreman-selinux-1.10.0-0.develop.################################# [ 50%]
Cleaning up / removing...
   2:foreman-selinux-1.10.0-0.develop.################################# [100%]
[root@smqa-r210-06-vm04 ~]# foreman-selinux-enable
```

domcleal:
> The moment docker policy is loaded then our rule gets added.

Hm, I don't think that's true because the docker_port_t type is actually defined our policy, not Docker's.  I'm pretty sure that we don't require docker-selinux for anything.  On EL6 would be the only possibility as there we use an existing macro, but this issue seems confined to EL7 for now.

domcleal:
> On EL6 would be the only possibility as there we use an existing macro, but this issue seems confined to EL7 for now

Ugh, misread, sorry.  It's ifndef, i.e. _not_ EL6.

domcleal:
Merged as caf0b6c, thanks @lzap.

PR #54, opened 2016-01-05
Fixes #12990 - allow reading of puppet symlinks

lzap:
It's a workaround for passenger.

Refpolicy is correct while symlinks are not in Fedora policy. Filed a PR for
them:

https://github.com/fedora-selinux/selinux-policy/pull/87


domcleal:
Merged as d0b68b3, thanks @lzap.

PR #55, opened 2016-01-11
Fixes #13041 - passenger var context fix

lzap:
If this is confirmed to work, I will go ahead and fix this upstream:

https://github.com/TresysTechnology/refpolicy-contrib/blob/master/passenger.fc

Change in Passenger 4.0 perhaps?

### DIFF
--- a/foreman-selinux-relabel
+++ b/foreman-selinux-relabel
@@ -15,7 +15,8 @@
   /etc/cron.d/foreman* \
   /usr/lib/ruby/gems/1.8/gems/passenger-* \
   /usr/lib64/gems/ruby/passenger-*/agents \
-  /usr/lib64/ruby/site_ruby/1.8/x86_64-linux/agents
+  /usr/lib64/ruby/site_ruby/1.8/x86_64-linux/agents \
+  /var/run/rubygem-passenger
 
 # relabel SCL mod_passenger and foreman plugins if SCL is found
 [ -d /opt/theforeman/tfm/ ] && /sbin/restorecon -ri $* \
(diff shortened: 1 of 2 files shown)


domcleal:
It's a patch in our package (https://github.com/theforeman/foreman-packaging/commit/a66b367702e9d0ee2b8e90dec8a8269153c4829d), which I think came from either Fedora or RHSCL packages.  The Fedora packages have since repackaged Passenger just as 'passenger' rather than 'rubygem-passenger', and now use /run/passenger again - so the patch is probably just needed here.

domcleal:
Looks fine, but could you also update the relabel script to ensure it fixes it on installation please?

<pre>
# ll -dZ /var/run/*passenger/                                                                                         
drwxr-xr-x. root root system_u:object_r:var_run_t:s0   /var/run/rubygem-passenger/
# rpm -Uvh ...
# ll -dZ /var/run/*passenger/                                                                                         
drwxr-xr-x. root root system_u:object_r:var_run_t:s0   /var/run/rubygem-passenger/
# restorecon -Rvvn /var/run/*passenger/
restorecon reset /var/run/rubygem-passenger context system_u:object_r:var_run_t:s0->system_u:object_r:passenger_var_run_t:s0
</pre>

lzap:
Good catch, added.

lzap:
If you have a minute or two, I'd love to get this in plus I found one issue I will be filing in a sec.

domcleal:
Merged as 86fa6e7, thanks @lzap.

PR #56, opened 2016-02-02
Refs #13502 - introduced has_docker definition

lzap:
This adds `has_docker` variable which is set when `docker.if` file is present.
This is not true on RHEL6 and RHEL 7.2 (as it was moved to separate
`docker-selinux` package there.

We used to have `distro_rhel6` there but we can now use this particular test
instead which is more accurate.

When this is compiled on RHEL 7.2 without the `docker-selinux` installed, it
does not fail. I will proceed with adding BuildRequire for our RPM package to
prevent this. As an indication, I've added special `defined_t` type which can
be used as a flag to easily find out if the `has_docker` support was compiled
in.

The patch also fixes few typos and adds some info messages in Makefile.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-.PHONY: clean remote-load
+.PHONY: clean remote-load load reload
 
 INSTPREFIX=
 VARIANT=targeted
@@ -8,10 +8,22 @@ MANCHAPTER=Foreman
 TMPDIR=local-tmp-foreman
 
 ifndef DISTRO
-$(error Set the DISTRO variable e.g. rhel7 or fedora21)
+$(error *** Set the DISTRO variable e.g. rhel7 or fedora21 ***)
+endif
+
+ifneq ("$(wildcard /usr/share/selinux/devel/include/*/docker.if)","")
+export M4PARAM += -D has_docker
+else
+ifneq ($(DISTRO),rhel6)
+$(error *** Interface docker.if not present, cannot continue ***)
+endif
 endif
 
 all: policies all-data
+
+load: \
+	foreman.pp.load.tmp \
+	foreman-proxy.pp.load.tmp
 
 policies: \
 	foreman.pp.bz2 \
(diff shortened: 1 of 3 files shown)


lzap:
Relevant PR: https://github.com/theforeman/foreman-packaging/pull/1006

Makefile
```diff
@@ -8,11 +8,21 @@ MANCHAPTER=Foreman
 TMPDIR=local-tmp-foreman
 
 ifndef DISTRO
-$(error Set the DISTRO variable e.g. rhel7 or fedora21)
+$(error *** Set the DISTRO variable e.g. rhel7 or fedora21 ***)
+endif
+
+ifneq ("$(wildcard /usr/share/selinux/devel/include/services/docker.if)","")
```
domcleal: In EL7.0 this is in contrib/docker.if, so it should probably handle either.

lzap: Good catch, also changed the warning to hard failure when docker is not present.

Makefile
```diff
@@ -8,11 +8,21 @@ MANCHAPTER=Foreman
...
 
 ifndef DISTRO
-$(error Set the DISTRO variable e.g. rhel7 or fedora21)
+$(error *** Set the DISTRO variable e.g. rhel7 or fedora21 ***)
+endif
+
+ifneq ("$(wildcard /usr/share/selinux/devel/include/*/docker.if)","")
+export M4PARAM += -D has_docker
+else
+$(error *** Interface docker.if not present, cannot continue ***)
```
domcleal, line 18: Will this error on EL6 without the file?

lzap, line 18: Yeah that was incorrect, fixed.

Makefile
```diff
@@ -8,11 +8,23 @@ MANCHAPTER=Foreman
...
+export M4PARAM += -D has_docker
+else
+ifneq ($(DISTRO),rhel6)
+$(error *** Interface docker.if not present, cannot continue ***)
+endif
 endif
 
 all: policies all-data
 
+load: \
```
domcleal, line 24: add to .PHONY

Makefile
```diff
@@ -62,12 +74,20 @@ foreman-proxy-selinux-disable: common/selinux-disable.sh
...
-	make -C ${TMPDIR} -f /usr/share/selinux/devel/Makefile NAME=${VARIANT} DISTRO=$(DISTRO)
+	$(MAKE) -C ${TMPDIR} -f /usr/share/selinux/devel/Makefile NAME=${VARIANT} DISTRO=$(DISTRO)
 	mv ${TMPDIR}/$@ .
 
+%.pp.tmp: %.pp
+	$(info ************ LOADING MODULE $< ************)
+	semodule -i $<
+	touch $@
+
+reload: clean load
```
domcleal, line 89: ditto, missing from .PHONY

Makefile
```diff
@@ -62,12 +74,20 @@ foreman-proxy-selinux-disable: common/selinux-disable.sh
...
+	cp $< ${<:.te=.fc} ${<:.te=.if} ${TMPDIR}/
 	sed -i 's/@@VERSION@@/${VERSION}/' ${TMPDIR}/*.te
-	make -C ${TMPDIR} -f /usr/share/selinux/devel/Makefile NAME=${VARIANT} DISTRO=$(DISTRO)
+	$(MAKE) -C ${TMPDIR} -f /usr/share/selinux/devel/Makefile NAME=${VARIANT} DISTRO=$(DISTRO)
 	mv ${TMPDIR}/$@ .
 
+%.pp.tmp: %.pp
+	$(info ************ LOADING MODULE $< ************)
+	semodule -i $<
+	touch $@
```
domcleal, line 87: These .pp.tmp files should be in .gitignore (perhaps ignore .pp.\* instead of specifics), and consider calling them .load or something as they're just markers?

domcleal:
Thanks, works well now - just a couple of minor inline comments above.  Please also change the commit message to "fixes".

lzap:
> Thanks, works well now - just a couple of minor inline comments above.  Please also change the commit message to "fixes".

Done, thanks.

## 

Later,
 Lukas #lzap Zapletal

domcleal:
Merged as eb34a92 and fixed the commit message, thanks @lzap.

PR #59, opened 2016-07-11
Fixes #15640 - added neutron port

lzap:
Port is defined in both RHEL6 and RHEL7.


dLobatog:
Merged as 264951f83dc8bbf845d71d7f4b16acdb6288961d, thanks @lzap !

PR #61, opened 2016-09-22
Fixes #16263 - fixed corenet_tcp_connect_neutron_port for 6.5

lzap:


ehelms:
:+1:

domcleal:
This will still fail to build on EL 6.5 as the `corenet_tcp_connect_neutron_port` macro is missing. `optional_policy` will only help _loading_ of policies on older versions. I don't think this fixes the issue.

lzap:
Yeah, I misunderstood your comment in the ticket, how about this? The quantum port is an alias so this should work in both old and new RHEL 6 versions: https://github.com/fedora-selinux/selinux-policy/commit/3d3549c0bdd84af83fe6d1f4f3c9379e65a8f73a

domcleal:
Merged, thanks @lzap.

PR #62, opened 2016-10-11
Fixes #16856 - added reading cert_t rules

lzap:


domcleal:
Should this policy not be in Katello, if it permits access to their certs? It probably shouldn't permit access to cert_t in a default installation.

The user also reports that this fix was insufficient in the ticket description.

lzap:
Not reading descriptions properly, a mistake I do all the time.

PR #66, opened 2017-01-27
Fixes #18284 - Remove docker_port_t definition

dLobatog:
docker_port_t is defined on container-selinux. We could make
foreman-selinux depend on container-selinux to be able to install both
modules together on the same computer.

https://github.com/projectatomic/container-selinux/blob/master/container.te#L77 is where the type is defined

### DIFF
--- a/foreman.te
+++ b/foreman.te
@@ -396,9 +396,6 @@ read_files_pattern(websockify_t, puppet_var_lib_t, puppet_var_lib_t)
 # Docker
 #
 
-type docker_port_t;
-corenet_port(docker_port_t)
-
 tunable_policy(`passenger_can_connect_docker_tcp',`
     allow passenger_t docker_port_t:tcp_socket name_connect;
 ')


foreman.te
```diff
@@ -396,9 +396,6 @@ read_files_pattern(websockify_t, puppet_var_lib_t, puppet_var_lib_t)
 # Docker
 #
 
-type docker_port_t;
-corenet_port(docker_port_t)
-
 tunable_policy(`passenger_can_connect_docker_tcp',`
     allow passenger_t docker_port_t:tcp_socket name_connect;
```
domcleal, line 400: Shouldn't this be optional policy if dependent on an additional module, and also conditional on `has_docker` too?

domcleal (changes requested):
The policy should probably be removed and maintained by the foreman_docker plugin, so as not to introduce any dependencies in Foreman on Docker or container policy.

foreman.te
lzap: Yes, this could help. But you made a good point on extracting this into separate package. That would be much cleaner.

lzap:
Alternative proposal, let's completely remove docker from our policy and we can only test for docker_port_t in the `enable` script, if present, we will enable it.

And for the upgrade path, once we remove our policy, the docker policy that can't load will load fine. This should work.

lzap:
@dLobatog are you planning to work on that? Dan also recommends to define the docker type this way: https://github.com/projectatomic/container-selinux/pull/26#issuecomment-284435882

lzap:
Replaced with https://github.com/theforeman/foreman-selinux/pull/68

PR #67, opened 2017-05-12
Fixes #9805 - fixed file_contexts: invalid context

lzap:
I was not able to reproduce but after multiple reports (see the associated
issue and bugzilla) and help from SELinux team of Red Hat we were able to
reveal a `dontaudit` missing rule that was causing this.

### DIFF
--- a/foreman.te
+++ b/foreman.te
@@ -265,6 +265,8 @@ optional_policy(`
         read_files_pattern(httpd_t, puppet_etc_t, puppet_etc_t)
         read_lnk_files_pattern(httpd_t, puppet_etc_t, puppet_etc_t)
         allow httpd_t puppet_port_t:tcp_socket name_connect;
+        # http://projects.theforeman.org/issues/9805
+        selinux_validate_context(passenger_t)
     ')
 ')
 


lzap:
@dLobatog I guess it's just you and me here. Thanks!

dLobatog (commented):
@lzap Is this still happening with Puppet 4? I'm trying on a foreman 1.15 production machine - as far as I understand the error should happen when the node.rb file is called after running `puppet agent -tv`, but I don't see it.

All in all, I don't fully understand what's causing the error in the issue (is this preventing just a warning or a denial?) or how to reproduce. If you can give me a way this should be reproduced that'd be the best - otherwise I'm not really sure what I'm merging. I'm happy to keep poking at it...

ares:
Does it have any side effect or does it do any harm? I'd like to to avoid merging things blindly, but if it's too hard to reproduce, I'd consider merging it in case it solves users problems. @lzap could any of the reporter test this before merge? Just so we know it fixes the original issue.

lzap:
I have never reproduced this thing, but if you read the monster BZ associated with this problem and amount of cases, you will see that this is a real issue and the workaround which was written by SELinux RH team for us works.

ares:
@dLobatog would it make sense to merge even without reproducing if we have confirmation from a user at https://bugzilla.redhat.com/show_bug.cgi?id=1202924#c48?

dLobatog (approved):

PR #68, opened 2017-09-01
Fixes #18284 - removed docker_t port

lzap:
This replaces https://github.com/theforeman/foreman-selinux/pull/66

So the idea here is to remove docker from our policy. We will not use `has_docker` flag anymore, we do not support RHEL6 anymore so this is useless - it is present in RHEL7 and Fedora.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -14,7 +14,7 @@ endif
 ifneq ("$(wildcard /usr/share/selinux/devel/include/*/docker.if)","")
 export M4PARAM += -D has_docker
 else ifneq ("$(wildcard /usr/share/selinux/devel/include/*/container.if)","")
-export M4PARAM += -D has_docker
+export M4PARAM += -D has_container
 else ifneq ($(DISTRO),rhel6)
 $(error *** Interface container.if or docker.if not present, cannot continue ***)
 endif
--- a/foreman.te
+++ b/foreman.te
@@ -20,19 +20,6 @@ policy_module(foreman, @@VERSION@@)
 
 #######################################
 #
-# Definitions
-#
-# This defines set of special (unused) types which are used for easier detection
-# what definitions the policy was compiled with. Use seinfo -tTYPE to find
-# particular flag.
-#
-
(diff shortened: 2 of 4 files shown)


Makefile
```diff
@@ -14,7 +14,7 @@ endif
 ifneq ("$(wildcard /usr/share/selinux/devel/include/*/docker.if)","")
 export M4PARAM += -D has_docker
 else ifneq ("$(wildcard /usr/share/selinux/devel/include/*/container.if)","")
-export M4PARAM += -D has_docker
+export M4PARAM += -D has_container
 else ifneq ($(DISTRO),rhel6)
 $(error *** Interface container.if or docker.if not present, cannot continue ***)
```
lzap, line 19: Flags has_docker or has_container are unused but we still error out when these are not present.

foreman.te
```diff
@@ -20,19 +20,6 @@ policy_module(foreman, @@VERSION@@)
...
-# what definitions the policy was compiled with. Use seinfo -tTYPE to find
-# particular flag.
-#
-
-ifdef(`has_docker', `
-    type foreman_has_docker_defined_t;
-')
-
-#######################################
-#
```
lzap, line 35: This was just a dummy type.

dLobatog (approved):
Thanks @lzap !!!

foreman.te
```diff
@@ -395,21 +382,18 @@ read_files_pattern(websockify_t, puppet_var_lib_t, puppet_var_lib_t)
...
+# Container / Docker
 #
 
-type docker_port_t;
-corenet_port(docker_port_t)
-
-tunable_policy(`passenger_can_connect_docker_tcp',`
-    allow passenger_t docker_port_t:tcp_socket name_connect;
+optional_policy(`
+    tunable_policy(`passenger_can_connect_docker_tcp',`
```
ares, line 389: @lzap this seems like a ` and `'` at the end of line are swapped

lzap: Sorry what? :)

PR #69, opened 2017-09-21
Fixes #21052 - compile policy with docker macros

lzap:
This should fix the compilation error.

SELinux is hard, different kernel, buildroot and target system combinations are hard to track.


dLobatog (approved):
Thanks @lzap , are you aware of any framework for testing this? I thought about just running the test_upgrade job with the newly compiled selinux policy, but there might be better options.

lzap:
I think there is nothing really, we would need to spin RHEL 7.0 - 7.4 and compile this there.

PR #70, opened 2017-09-21
Refs #21052 - workaround missing docker macro

lzap:
The macro is not available in our build root, we have older version of RHEL there. This should do it, unless spc_t is also not available.

### DIFF
--- a/foreman.te
+++ b/foreman.te
@@ -402,7 +402,14 @@ optional_policy(`
             container_spc_stream_connect(passenger_t)
         ')
         ifdef(`has_docker', `
-            docker_spc_stream_connect(passenger_t)
+            #docker_spc_stream_connect(passenger_t)
+            gen_require(`
+                type spc_t, spc_var_run_t;
+                attribute pidfile;
+            ')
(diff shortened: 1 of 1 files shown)


foreman.te
```diff
@@ -402,7 +402,13 @@ optional_policy(`
             container_spc_stream_connect(passenger_t)
         ')
         ifdef(`has_docker', `
-            docker_spc_stream_connect(passenger_t)
+            #docker_spc_stream_connect(passenger_t)
+            gen_require(`
+                type spc_t, spc_var_run_t;
+            ')
+            files_search_pids(passenger_t)
+            files_write_all_pid_sockets(passenger_t)
```
dLobatog: This line doesn't seem to be available on Koji

dLobatog (changes requested):
@lzap Thanks for the quick fix, apparently this doesn't move forward on Koji either - http://koji.katello.org/kojifiles/work/tasks/259/30259/build.log

dLobatog:
Already cp-ed to 1.16-stable, has to be cp-ed to 1.15

PR #72, opened 2017-11-06
Refs #18284 - added foreman_container_port_t

lzap:
During refactoring in the referenced ticket, I deleted `docker_port_t` type which was conflicting with container-selinux package. Unfortunately it turned out this port cannot be used to create allow rule for TCP connections, it is defined but there is no port number labeled with it.

This patch fixes this by introducing new type and assigning ports to it, but only whent they are not labeled yet. I assume the port assignements will be done in upstream and this will be rebased into RHEL later.

This patch also drops a legacy RHEL6 code from enable script and removes `foreman_can_connect_docker_tcp` boolean which is no longer relevant (we can always connect it as long as the port is assigned, unassign to disallow this). It also removes the other boolean from "docker" to "container" for the future.

### DIFF
--- a/foreman-selinux-enable
+++ b/foreman-selinux-enable
@@ -6,10 +6,6 @@ TMP_EXEC_AFTER=$(mktemp -t foreman-selinux-enable.XXXXX)
 TMP_PORTS=$(mktemp -t foreman-selinux-enable.XXXXX)
 trap "rm -rf '$TMP_EXEC_BEFORE' '$TMP_EXEC_AFTER' '$TMP_PORTS'" EXIT INT TERM
 
-is_redhat_6() {
-  [[ $(rpm -q --whatprovides redhat-release --qf '%{version}') == 6* ]]
-}
-
 # Load or upgrade foreman policy and set booleans.
 #
 # Dependant booleans must be managed in a separate transaction.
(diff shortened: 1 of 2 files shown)


dLobatog (approved):
Thanks @lzap ! I will try to build it in Koji and if it works, I'll merge - trying to avoid the havoc we went through last time 😄

foreman-selinux-enable
```diff
@@ -6,10 +6,6 @@ TMP_EXEC_AFTER=$(mktemp -t foreman-selinux-enable.XXXXX)
 TMP_PORTS=$(mktemp -t foreman-selinux-enable.XXXXX)
 trap "rm -rf '$TMP_EXEC_BEFORE' '$TMP_EXEC_AFTER' '$TMP_PORTS'" EXIT INT TERM
 
-is_redhat_6() {
-  [[ $(rpm -q --whatprovides redhat-release --qf '%{version}') == 6* ]]
-}
-
```
lzap, line 12: This is not useful anymore.

foreman-selinux-enable
```diff
@@ -21,16 +17,16 @@ do
   if /usr/sbin/semodule -s $selinuxvariant -l >/dev/null; then
     /usr/sbin/semanage port -E > $TMP_PORTS
 
-    # Remove previously defined docker_port_t
-    # (this can be removed in future release)
+    # Remove previously defined conflicting docker_port_t (this can be removed in future release)
     grep docker_port_t $TMP_PORTS | sed s/-a/-d/g >> $TMP_EXEC_BEFORE
```
lzap, line 21: We need to keep removing the conflicting `docker_port_t` in BEFORE block.

foreman-selinux-enable
```diff
@@ -21,16 +17,16 @@ do
...
 
-    echo "boolean -m --on httpd_setrlimit" >> $TMP_EXEC_AFTER
+    # Assign docker/container port only and only if it's undefined
+    grep -qE 'tcp 2375' $TMP_PORTS || \
+      echo "port -a -t foreman_container_port_t -p tcp 2375" >> $TMP_EXEC_AFTER
+    grep -qE 'tcp 2376' $TMP_PORTS || \
+      echo "port -a -t foreman_container_port_t -p tcp 2376" >> $TMP_EXEC_AFTER
```
lzap, line 27: And in AFTER block we insert port numbers to `foreman_container_port_t` only when they are not present. I could have used a single `2375-2376` line but then grepping would be too complex.

foreman-selinux-enable
```diff
@@ -21,16 +17,16 @@ do
...
 
-    if is_redhat_6; then
-      grep -q foreman_osapi_compute_port_t $TMP_PORTS || \
-        echo "port -a -t foreman_osapi_compute_port_t -p tcp 8774" >> $TMP_EXEC_AFTER
-    fi
+    echo "boolean -m --on httpd_setrlimit" >> $TMP_EXEC_AFTER
```
lzap, line 29: This was just moved down here.

foreman.te
```diff
@@ -64,17 +64,10 @@ gen_tunable(passenger_can_connect_all, false)
 
 ## <desc>
 ## <p>
-## Determine whether passenger can connect to Docker via TCP
-## </p>
-## </desc>
-gen_tunable(passenger_can_connect_docker_tcp, true)
-
-## <desc>
-## <p>
```
lzap, line 73: No longer needed, we allow connections to all ports added. User can remove them.

foreman.te
```diff
@@ -385,23 +381,16 @@ read_files_pattern(websockify_t, puppet_var_lib_t, puppet_var_lib_t)
...
+    tunable_policy(`passenger_can_connect_container_unix',`
         ifdef(`has_container', `
+            container_stream_connect(passenger_t)
             container_spc_stream_connect(passenger_t)
         ')
         ifdef(`has_docker', `
+            docker_stream_connect(passenger_t)
             #docker_spc_stream_connect(passenger_t)
             gen_require(`
                 type spc_t, spc_var_run_t;
```
lzap, line 396: Renamed the boolean and added missing `*_stream_connect` - this was actually a bug, it was in the `tcp` block but in does belong here.

dLobatog:
http://koji.katello.org/koji/taskinfo?taskID=48675 building...

lzap:
Any news? Thanks.

dLobatog (approved):
Thanks @lzap

PR #73, opened 2017-12-06
Fixes #21887 - allow connecting to puppet master

lzap:
This showed up in 1.15 for some reason.

### DIFF
--- a/foreman.te
+++ b/foreman.te
@@ -85,6 +85,20 @@ gen_tunable(passenger_can_spawn_ssh, true)
 
 ## <desc>
 ## <p>
+## Determine whether passenger can connect to http_port_t when preloading app.
+## </p>
+## </desc>
+gen_tunable(passenger_can_connect_http, true)
+
+## <desc>
+## <p>
+## Determine whether passenger can connect to puppet_port_t when preloading app.
+## </p>
+## </desc>
+gen_tunable(passenger_can_connect_puppetmaster, true)
+
+## <desc>
+## <p>
 ## Determine whether passenger can connect to local or remote libvirt.
 ## </p>
 ## </desc>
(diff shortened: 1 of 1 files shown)


lzap:
Here @ares @ekohl I am allowing this port, but was not able to reproduce with default install. If you confirm this is non-standard, I will make the boolean to be false by default so users need to explicitly enable this.

ekohl (approved):
@lzap I'm guessing that your 1.15 install was with puppetserver (standalone java process) rather than the deprecated puppet-server (ruby via passenger).

lzap:
Can you please explain in a little bit more detail? This does not tell me why this TCP port was unused in our old Puppet versions and now is required, which bit of code does the request please? Thanks.

ekohl:
https://github.com/theforeman/puppet-puppet/commit/62fa011bf4e087b21008c508ba864ebd293c4eee#diff-2b7703dcdb4485ab28414b14bab9ab3b introduced the preloading. It was first included in foreman-installer 1.12. Since this is optional I suspect most people didn't notice the denial or don't use selinux or use puppetserver.

lzap:
Thanks, so if I understand correctly, this is actually Passenger preloading the HTTP(S) endpoint, not Foreman itself. Then the patch is correct.

ekohl:
@lzap I asked Peter to confirm that for me in the BZ but that is what I suspect yes.

lzap:
Thanks, this put some light on this, I reworded some rules, we had an old one which I did not understand, now it's more clear. I made this very explicit using two boolean flags now.

foreman.te
```diff
@@ -83,6 +83,20 @@ gen_tunable(passenger_can_connect_openstack, true)
 ## </desc>
 gen_tunable(passenger_can_spawn_ssh, true)
 
+## <desc>
+## <p>
+## Determine whether passenger can connect to itself (HTTPS) when preloading app.
```
ekohl: Mixing HTTPS and ```_http``` is a bit confusing. Is the http port on port 80, 443 or both?

foreman.te
```diff
@@ -83,6 +83,20 @@ gen_tunable(passenger_can_connect_openstack, true)
...
+## </p>
+## </desc>
+gen_tunable(passenger_can_connect_http, true)
+
+## <desc>
+## <p>
+## Determine whether passenger can connect to puppet master (HTTPS) when preloading app.
```
ekohl: ```the puppet master port``` and ```the app```. Also not sure if HTTPS here should be mentioned since it's really just a TCP connection that's allowed. What protocol they use isn't really checked by SELinux.

ekohl (changes requested):

foreman.te
lzap: You are right, in both cases port types are abstract things, they refer to 0-N port numbers. Changed.

ekohl (approved):

lzap:
@dLobatog this is good to go, thanks guys.

dLobatog (approved):
Thanks @ekohl @lzap !

dLobatog:
I'll cp for 1.16.1

lzap:
Thanks a bunch.

PR #74, opened 2017-12-19
Fixes #22028 - allowed searching attributes

lzap:
I don't know why passenger sniffs attributes around whole filesystem, I cannot reproduce, but this is from our dogfooding server.


lzap:
Explained in https://projects.theforeman.org/issues/13357 and this is not worth adding.

PR #75, opened 2017-12-19
Fixes #19053 - allow logrotate to send signals

lzap:
This is a temporary workaround to fix logrotate denials, multiple users reported this, both smart-proxy and smart-proxy dynflow core has the same issue. BZ that will fix this is RHEL 7.6 most likely, we need a solution today.

https://bugzilla.redhat.com/show_bug.cgi?id=1527522

### DIFF
--- a/foreman-proxy.te
+++ b/foreman-proxy.te
@@ -153,6 +153,12 @@ ifdef(`distro_rhel6', `
     miscfiles_read_generic_certs(foreman_proxy_t)
 ')
 
+# Logrotate in RHEL7 does not allow signals, RHBZ#1527522
+gen_require(`
+    type logrotate_t;
+')
+systemd_config_generic_services(logrotate_t)
+
 # generic support for plugins
 # executes sudo
 optional_policy(`
--- a/foreman.te
+++ b/foreman.te
@@ -231,6 +231,12 @@ miscfiles_manage_generic_cert_files(passenger_t)
 # /tmp/passenger-native-support-XXXXXX/passenger_native_support.so
 can_exec(passenger_t, passenger_tmp_t)
 
+# Logrotate in RHEL7 does not allow signals, RHBZ#1527522
+gen_require(`
+    type logrotate_t;
+')
+systemd_config_generic_services(logrotate_t)
+
 optional_policy(`
     tunable_policy(`passenger_run_foreman', `
         admin_pattern(httpd_t, foreman_lib_t, foreman_lib_t)


yoctozepto:
I believe this is a too broad change. It would be better to introduce a new type for foreman-proxy service.

lzap:
After futher discussion, @yoctozepto revealed that rotated is already allowed to send arbitrary signals anyway, so this is no longer a concern.

yoctozepto:
Yes, I confirm. It is highly unlikely that this change would make SELinux protections any worse. The bottom line is that one should audit logrotate scripts thoroughly.

I would fix the comment in code though. Maybe:
`# In RHEL7.4 and older logrotate is not allowed to send signals via systemd, RHBZ#1527522`

Also, only foreman-proxy is affected, not foreman itself, as it does not require a reload.

tbrisker:
This PR has not seen any activity in over a year, I'm assuming it is no longer relevant and closing, please reopen if I'm wrong.

lzap:
@tbrisker this is still a valid PR it's just nobody cared to look into this. Please reopen, I don't have that button available.

lzap:
Poking @ekohl or @xprazak2 with this one, explanation on why this must be there is in the BZ, but in short logrotate script in Foreman performs `systemctl xxxx foreman.service` and base policy will not allow that. So we need to do this.

lzap:
Rebased.

ehelms:
@ekohl or @xprazak2 Any comments for @lzap ? otherwise looks sane, and I'll merge in a few days

foreman-proxy.te
```diff
@@ -153,6 +153,13 @@ ifdef(`distro_rhel6', `
     miscfiles_read_generic_certs(foreman_proxy_t)
 ')
 
+# Logrotate in RHEL7 does not allow signals, RHBZ#1527522
+gen_require(`
+    type logrotate_t;
+    type systemd_unit_file_t;
+')
+allow logrotate_t systemd_unit_file_t:service stop;
```
ekohl: I think it should only send signals which AFIAK is `kill`, but never stop. Is this part of `stop` in selinux?

lzap: I am actually dropping this code and changing it to the macro that was introduced in RHEL 7.7.

lzap:
Rebased, now the code is exactly the same as what goes into RHEL 7.7.

foreman.te
```diff
@@ -231,6 +231,13 @@ miscfiles_manage_generic_cert_files(passenger_t)
 # /tmp/passenger-native-support-XXXXXX/passenger_native_support.so
 can_exec(passenger_t, passenger_tmp_t)
 
+# Logrotate in RHEL7 does not allow signals, RHBZ#1527522
+gen_require(`
+    type logrotate_t;
+    type systemd_unit_file_t;
+')
+allow logrotate_t systemd_unit_file_t:service stop;
```
ekohl: For my understanding, is stop just generic for a signal?

ekohl (commented):
Since CentOS 7.7 is not yet released, will this be an issue?

foreman.te
lzap: Oh this is a left over. It must be same as in `foreman-proxy.te` thanks.

lzap:
The purpose of this patch is actually to pull the fix from 7.7 into our repos. Once users upgrade to 7.7 there will be two same rules in the policy but that does not error out - it's something that works. We can remove those workarounds later on (we actually carry many of them - e.g. for puppet etc).

foreman-proxy.te
tbrisker: > I am actually dropping this code and changing it to the macro that was introduced in RHEL 7.7.

Is this macro available in EL < 7.7?

lzap:
Macro `systemd_config_generic_services` is present from 7.0.

tbrisker (approved):
Thanks @lzap !

PR #79, opened 2018-04-05
Fixes #23134 - support for plugin hooks

lzap:
We need to allow to katello-selinux to hook onto foreman-selinux-enable (disable, relabel) scripts easily.

This can be done via /usr/libexec/foreman-selinux/enable.d/*.after.sh (and before.sh) scripts executed before or after scripts.

Packaging PR will follow once this is merged to get that directory created.

### DIFF
--- a/foreman-proxy-selinux-relabel
+++ b/foreman-proxy-selinux-relabel
@@ -1,5 +1,10 @@
 #!/bin/bash
 
+LIBEXEC_DIR=/usr/libexec/foreman-selinux
+
+# Run hooks
+find ${LIBEXEC_DIR} -name \*-before-relabel.sh -type f -executable -exec /usr/bin/bash '{}' \;
+
 /sbin/restorecon -ri $* \
   /usr/share/foreman-proxy/bin/smart-proxy \
   /etc/foreman-proxy \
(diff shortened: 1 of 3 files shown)


adamruzicka (approved):
Looks good under the assumption there can't be spaces in `$LIBEXEC_DIR` or in hook names

lzap:
You want me to workaround it? I'd be very surprised there was a libexec executable with space, this is not MacOS :-)

adamruzicka:
Working around it should be quite easy and I'd rather stay on the safe side, just to be sure.

lzap:
Ok done.

foreman-proxy-selinux-relabel
```diff
@@ -8,4 +15,9 @@
   /var/log/foreman-proxy \
   /var/run/foreman-proxy
 
+# Run hooks
+for HOOK in "$LIBEXEC_DIR/*-after-relabel.sh"; do
```
adamruzicka: The escaping needs to be done in a slightly different way, see [here](https://gist.github.com/adamruzicka/e15307a67309dd46250adc4c3fbe5cea)

lzap: Hmmm good catch, I have tested this incorrectly.

lzap:
I actually could not get your example work, but found different solution via find, no shell involved.

adamruzicka (approved):
Looks good to me

PR #80, opened 2018-05-17
Fixes #23619 - disable script removes foreman ports

lzap:
The disable script was not removing all definition, it now catches all "foreman_*_port_t" ports properly. This made reload of policy impossible.


lzap:
This is a quick merge - someone? @ares

adamruzicka (approved):
Not much to review here, ACK

PR #81, opened 2018-05-25
Fixes #23700 - passenger/http can manage symlinks

lzap:
Foreman Templates plugin needs this, ships with Satellite 6.3

### DIFF
--- a/foreman.te
+++ b/foreman.te
@@ -220,6 +220,9 @@ optional_policy(`
         # Allow Foreman to do DNS queries via Ruby stdlib net package
         # http://projects.theforeman.org/issues/8030
         corenet_udp_bind_generic_port(passenger_t)
+
+        # Templates plugin manages symlinks in tmp dir
+        manage_lnk_files_pattern(passenger_t, passenger_tmp_t, passenger_tmp_t)
     ')
 ')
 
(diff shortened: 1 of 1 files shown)


foreman.te
```diff
@@ -236,7 +239,11 @@ optional_policy(`
 ')
 
 tunable_policy(`httpd_run_foreman', `
-    allow httpd_t passenger_tmp_t:sock_file write;
+    manage_dirs_pattern(httpd_t, passenger_tmp_t, passenger_tmp_t)
+    manage_files_pattern(httpd_t, passenger_tmp_t, passenger_tmp_t)
+    manage_sock_files_pattern(httpd_t, passenger_tmp_t, passenger_tmp_t)
```
adamruzicka, line 244: I assume `manage_sock_files_pattern` replaces `allow httpd_t passenger_tmp_t:sock_file write;`, is that correct?

adamruzicka (approved):
Fixes the issue for me, left one comment inline just to check I'm interpreting the macros correctly.

foreman.te
lzap: Exactly, see here: https://github.com/fedora-selinux/selinux-policy/blob/rawhide/policy/support/file_patterns.spt#L373-L376

Grep the repo further to find `policy/support/obj_perm_sets.spt` file holding another macro with all the actions which includes the `write` operation. Just keep in mind when looking up macros that we target CentOS7 and git repo for that is internal (or use centos7 SRPM instead). There is lot of development already.

lzap:
We have another report the patch works also with Satellite 6.3.

foreman.te
lzap: To further explain, passenger_t domain is already allowed all above, but httpd_t is not. We want to keep both on par when the foreman_run_httpd flag is turned on. It was a bit behind.

dgoetz:
Verified it successfully on Foreman 1.17.

adamruzicka:
@ares (or anyone with the right permissions) Could you click the green button for us?

lzap:
Thanks for your reviews, appreciated! @adamruzicka ++

PR #82, opened 2018-06-28
Fixes #11572 - add support for spool files

dgoetz:
refs #11572


lzap:
Nicely done, merging untested because our **proxy** policy is not enabled by default (it needs much more love).

PR #86, opened 2018-09-15
Fixes #24952 - Check if the directory exists before executing

ekohl:
Otherwise you get warnings:

    find: '/usr/libexec/foreman-selinux': No such file or directory


adamruzicka (approved):
LGTM

lzap (approved):
I intentionally did not do this since RPM package delivers it, but this can be problem when testing this without RPM package installed.

lzap:
Thanks merged.

PR #87, opened 2019-01-07
Fixes #25783: Allow websockify to read puppet_etc_t

bmeekhof:


lzap (approved):
Ideally puppet policy should create a separate context for those (sensitive) files but this works. Thanks!

PR #88, opened 2019-04-03
Fixes #16273 - allow execmem

lzap:
FFI library callback feature executes mapped memory of temporary files. This
sounds dangerous, however FFI authors do not think this is a problem, they did
not find a workaround in the FFI code as a good option and they recommend to
allow this in SELinux in the official documentation:

* https://bitbucket.org/cffi/cffi/issues/231
* https://cffi.readthedocs.io/en/latest/using.html#callbacks

This patch enables execmem both for passenger and smart proxy process as there
is no other good solution to this problem.

### DIFF
--- a/foreman-proxy.te
+++ b/foreman-proxy.te
@@ -104,6 +104,9 @@ files_spool_filetrans(foreman_proxy_t, foreman_proxy_spool_t, { dir file })
 # starting via /bin/env
 corecmd_read_bin_symlinks(foreman_proxy_t)
 
+# FFI library callbacks - see RM#26520 for more info
+allow foreman_proxy_t self:process execmem;
+
 # ruby runtime
 corecmd_search_bin(foreman_proxy_t)
 corecmd_exec_bin(foreman_proxy_t)


ekohl:
This is only needed by Zest which is only used in Katello. Perhaps a candidate for katello-selinux?

lzap:
Do you insist? I'd rather be precausious and since we've run into this in the past (we actually had execmem for some time in core) I am for having this in core.

ekohl:
You typically want to allow as little as possible and execmem sounds scary. If core doesn't use it, it'd be good to avoid it which was my reason to move it to katello-selinux.

ehelms:
I'd love to see katello-selinux merged here so there is only a single selinux policy across the ecosystem if that helps.

ekohl:
I have considered a katello selinux boolean in this package so we have one package, but only allow the minimum needed.

lzap:
I've split this into foreman-proxy part (here) and katello-selinux (passenger): https://github.com/Katello/katello-selinux/pull/19

lzap:
This is ready for merge, @tbrisker wanted us to unblock nightly. After this is merged I will go ahead and push updated packages.

ekohl (approved):
Not happy this is needed, but I guess we don't really have a choice here.

tbrisker (approved):

lzap:
Nightly with the patch is out: http://yum.theforeman.org/nightly/el7/x86_64/foreman-selinux-1.22.0-0.1.develop.20190403150122git390a568.el7.noarch.rpm

Thanks.

tbrisker:
@lzap iirc you said we should add an upgrade warning regarding this to the manual, care to open a PR against nightly with it?

PR #90, opened 2019-05-08
drop rhel6 vs rhel7 ifdefs, we don't support rhel6 anymore

evgeni:


evgeni:
rebased

lzap (approved):
Thanks merging.

PR #91, opened 2019-05-08
Fixes #26762 - move auth_read_passwd ouside the tunable_policy

evgeni:
auth_read_passwd in EL8 (and modern Fedora) includes an optional
statement, which cannot be inside a tunable policy.
This moves it outside the tunable, making it always enabled. Not
optimal, but it makes the policy work across all distros.


lzap:
Thanks it's fine.

PR #93, opened 2019-06-04
Fixes #26951 - allow execmem in passenger too

lzap:
We had this rule, then we removed it but it's back. Let's add it back.

Similar issue was: https://github.com/theforeman/foreman-selinux/pull/88


ehelms (approved):

lzap:
@ehelms please merge I guess it's just me and @xprazak2 with push rights here.

PR #94, opened 2019-10-29
Fixes #29115 - policy for foreman-cockpit service

lzap:
Policy for Foreman Cockpit integration. Foreman connects to Cockpit running on custom port 19090 (launched via foreman-cockpit.service provided by Remote Execution) for authorization. Port 9090 could not be used as it's in used by Smart Proxy.

This works be using existing `websm_port_t` which is used by Cockpit in base policy and assigning the free port number 19090 to it. The passenger domain already has a rule to connect to websm port (9090) because that's what Smart Proxy uses, so I only left a comment there.

https://github.com/theforeman/foreman_remote_execution/pull/441/files

### DIFF
--- a/foreman-selinux-enable
+++ b/foreman-selinux-enable
@@ -37,11 +37,14 @@ do
     # Create port list cache
     /usr/sbin/semanage port -E > $TMP_PORTS
 
-    # Assign docker/container port only and only if it's undefined
+    # Assign foreman custom ports
     grep -qE 'tcp 2375' $TMP_PORTS || \
       echo "port -a -t foreman_container_port_t -p tcp 2375" >> $TMP_EXEC_AFTER
     grep -qE 'tcp 2376' $TMP_PORTS || \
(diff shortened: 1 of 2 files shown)


tbrisker (approved):
Thanks @lzap !

tbrisker:
1.24 - 332faa1

lzap:
For the record, the RM number was supposed to be 28115

tbrisker:
force pushed with correct message:
develop - d9ab83762dbd5a8f3034a373cc333682050647f7
1.24 - 252ac88

ekohl:
> The passenger domain already has a rule to connect to websm port (9090) because that's what Smart Proxy uses

Technically it's slightly more complicated. In vanilla Foreman port 9090 is also the default cockpit port. There is the explicit choice to never reuse that instance and always spawn a standalone instance. You are correct that in the Katello scenario it would also conflict with the Smart Proxy.

PR #95, opened 2019-11-18
Fixes #28252 - rules for foreman cockpit session

lzap:
Fixes a denial when switching from Foreman to Cockpit. Tested with Satellite 6.7 snap 2, however this should work with Foreman 1.24 too.

https://bugzilla.redhat.com/show_bug.cgi?id=1770999#c4

### DIFF
--- a/foreman.fc
+++ b/foreman.fc
@@ -51,6 +51,11 @@
 /usr/share/gems/gems/passenger-.*/helper-scripts/prespawn -- gen_context(system_u:object_r:passenger_exec_t,s0)
 /usr/lib6?4?/gems/(exts|ruby)/passenger-.*/agents/.* -- gen_context(system_u:object_r:passenger_exec_t,s0)
 
+# Foreman Remote Execution
+
+/usr/sbin/foreman-cockpit-session       gen_context(system_u:object_r:cockpit_session_exec_t,s0)
+/usr/share/gems/gems/foreman_remote_execution-.*/extra/cockpit/foreman-cockpit-session -- gen_context(system_u:object_r:cockpit_session_exec_t,s0)
+
 # Foreman Hooks plugin
 
 /usr/share/foreman/config/hooks(/.*)?   gen_context(system_u:object_r:foreman_hook_t,s0)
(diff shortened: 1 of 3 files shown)


foreman.fc
```diff
@@ -51,6 +51,11 @@
 /usr/share/gems/gems/passenger-.*/helper-scripts/prespawn -- gen_context(system_u:object_r:passenger_exec_t,s0)
 /usr/lib6?4?/gems/(exts|ruby)/passenger-.*/agents/.* -- gen_context(system_u:object_r:passenger_exec_t,s0)
 
+# Foreman Remote Execution
+
+/usr/sbin/foreman-cockpit-session       gen_context(system_u:object_r:cockpit_session_exec_t,s0)
+/usr/share/gems/gems/foreman_remote_execution-.*/extra/cockpit/foreman-cockpit-session -- gen_context(system_u:object_r:cockpit_session_exec_t,s0)
```
ekohl, line 57: This makes me wonder if we should move the file in packaging rather than making a symlink to it in `/usr/sbin`.

foreman.te
```diff
@@ -350,6 +349,28 @@ optional_policy(`
     ')
 ')
 
+#######################################
+#
+# Remote Execution
+#
+
+# Allow Foreman to connect to Foreman Proxy on port 9090 (Katello)
```
ekohl: The placing here implies this is REX but it's core Foreman functionality.

lzap: Good catch, I've moved this block accidentally actually.

foreman.fc
lzap: I don't mind but this is a typical pattern we use across many packages.

lzap:
Rebased, if you can smash the merge button we need this to unblock QA automation. This is tested end-to-end, I guarantee it works fine don't test it.

foreman.fc
ekohl: Given this reflects the current package, I'm fine with merging it.

ekohl (approved):

lzap:
Thank you, CP 1.24: https://github.com/theforeman/foreman-selinux/pull/96

PR #100, opened 2020-05-15
Fixes #29778 - puma policy, passenger optional

lzap:
This drops all unnecessary rules around weird stuff that passenger does (complipation of ruby extensions, inprocess monitoring, other woes). Then it introduces new domain `foreman_rails_t` and transition into the domain and then changes all still relevant rules.

Do not test or review yet, it currently does not compile. Need to ask SELinux experts about this new CIL errors.

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,12 +19,19 @@ Make sure you provide the correct distribution. Possible values are:
 
 There's a target to do compile and load policy on a remote system via ssh:
 
-    make remote-load host=my.host.lan distro=rhel7 name=foreman
+    make remote-load HOST=foreman.example.com DISTRO=rhel7 NAME=foreman
 
 Often it is necessary to relabel relevant files and directories:
 
     ssh my.host.lan
     my# ./foreman-selinux-relabel
+
+Debugging CIL
+-------------
+
+From time to time, SELinux spills out a cryptic error. To generate CLI source from te/pp file, do:
+
+    cat foreman.pp | /usr/libexec/selinux/hll/pp > /tmp/foreman.cil
 
 License
 -------
--- a/foreman-selinux-relabel
+++ b/foreman-selinux-relabel
@@ -10,9 +10,11 @@
   /etc/puppet/node.rb \
   /etc/puppetlabs/puppet/node.rb \
   /etc/sysconfig/foreman* \
-  /etc/rc.d/init.d/foreman* \
+  /usr/lib/systemd/system/foreman* \
+  /etc/systemd/system/foreman* \
   /etc/logrotate.d/foreman* \
   /etc/cron.d/foreman* \
+  /usr/libexec/foreman \
   /usr/lib{64,}/ruby/gems/1.8/gems/passenger-* \
   /usr/lib{64,}/gems/ruby/passenger-*/agents \
   /usr/lib{64,}/ruby/site_ruby/1.8/x86_64-linux/agents \
--- a/foreman.fc
+++ b/foreman.fc
@@ -16,21 +16,26 @@
 # this program. If not, see http://www.gnu.org/licenses/.
 #
 
-# Foreman file contexts
+# Foreman and Ruby on Rails
 
-/etc/foreman(/.*)?                      gen_context(system_u:object_r:foreman_config_t,s0)
-/etc/puppet/node.rb                 --  gen_context(system_u:object_r:foreman_enc_t, s0)
-/etc/puppetlabs/puppet/node.rb      --  gen_context(system_u:object_r:foreman_enc_t, s0)
+/usr/lib/systemd/system/foreman.*               --  gen_context(system_u:object_r:foreman_rails_unit_file_t,s0)
+/etc/systemd/system/foreman.*                   --  gen_context(system_u:object_r:foreman_rails_unit_file_t,s0)
 
-/var/lib/foreman(/.*)?                  gen_context(system_u:object_r:foreman_lib_t,s0)
-/var/lib/foreman/db/(.*.sqlite3)?       gen_context(system_u:object_r:foreman_db_t,s0)
+/etc/foreman(/.*)?                                  gen_context(system_u:object_r:foreman_config_t,s0)
+/etc/puppet/node.rb                             --  gen_context(system_u:object_r:foreman_enc_t, s0)
+/etc/puppetlabs/puppet/node.rb                  --  gen_context(system_u:object_r:foreman_enc_t, s0)
 
-/var/log/foreman(/.*)?                  gen_context(system_u:object_r:foreman_log_t,s0)
+/var/lib/foreman(/.*)?                              gen_context(system_u:object_r:foreman_lib_t,s0)
 
-/var/run/foreman(/.*)?                  gen_context(system_u:object_r:foreman_var_run_t,s0)
+/var/log/foreman(/.*)?                              gen_context(system_u:object_r:foreman_log_t,s0)
 
-/usr/share/foreman/.ssh(/.*)?           gen_context(system_u:object_r:ssh_home_t,s0)
-/usr/share/foreman/extras/noVNC/websockify\.py gen_context(system_u:object_r:websockify_exec_t,s0)
+/var/run/foreman(/.*)?                              gen_context(system_u:object_r:foreman_var_run_t,s0)
+
+/usr/share/foreman/.ssh(/.*)?                       gen_context(system_u:object_r:ssh_home_t,s0)
+/usr/share/foreman/extras/noVNC/websockify\.py      gen_context(system_u:object_r:websockify_exec_t,s0)
+
+/usr/share/foreman/bin/rails                    --  gen_context(system_u:object_r:foreman_rails_exec_t,s0)
+/usr/libexec/foreman/.*-selinux                 --  gen_context(system_u:object_r:foreman_rails_exec_t,s0)
 
 # Passenger non-SCL file contexts
 
--- a/foreman.te
+++ b/foreman.te
@@ -22,45 +22,35 @@ policy_module(foreman, @@VERSION@@)
 #
 # Declarations
 #
-# Note: Some mod_passanger 4.0+ processes are running under httpd_t domain,
-# therefore we need to keep two domains: passenger_t and httpd_t.
-#
 
 ## <desc>
 ## <p>
-## Allow passenger to run foreman.
+## Allow Ruby on Rails to run foreman (deprecated).
 ## </p>
 ## </desc>
-gen_tunable(passenger_run_foreman, true)
-
-## <desc>
-### <p>
-### Allow passenger to run the puppetmaster.
-### </p>
-### </desc>
-gen_tunable(passenger_run_puppetmaster, true)
+gen_tunable(passenger_run_foreman, false)
 
 ## <desc>
 ## <p>
-## Allow passenger (httpd_t domain) to run foreman.
+## Determine whether Ruby on Rails can listen/bind any port.
 ## </p>
 ## </desc>
-gen_tunable(httpd_run_foreman, true)
+gen_tunable(foreman_rails_can_listen_any, false)
 
 ## <desc>
 ## <p>
-## Determine whether passenger can connect to LDAP ports.
+## Determine whether Ruby on Rails can connect to LDAP ports.
 ## </p>
 ## </desc>
-gen_tunable(passenger_can_connect_ldap, true)
+gen_tunable(foreman_rails_can_connect_ldap, true)
 
 ## <desc>
 ## <p>
-## Determine whether passenger can connect to TCP ports
+## Determine whether Ruby on Rails can connect to TCP ports
 ## other than specified in the policy.
 ## </p>
 ## </desc>
-gen_tunable(passenger_can_connect_all, false)
+gen_tunable(foreman_rails_can_connect_all, false)
 
 ## <desc>
 ## <p>
@@ -71,55 +61,68 @@ gen_tunable(websockify_can_connect_all, false)
 
 ## <desc>
 ## <p>
-## Determine whether passenger can connect to Docker via local socket
+## Determine whether Ruby on Rails can connect to Docker via local socket
 ## </p>
 ## </desc>
-gen_tunable(passenger_can_connect_container_unix, true)
+gen_tunable(foreman_rails_can_connect_container_unix, true)
 
 ## <desc>
 ## <p>
-## Determine whether passenger can connect to OpenStack
+## Determine whether Ruby on Rails can connect to OpenStack
 ## </p>
 ## </desc>
-gen_tunable(passenger_can_connect_openstack, true)
+gen_tunable(foreman_rails_can_connect_openstack, true)
 
 ## <desc>
 ## <p>
-## Determine whether passenger can execute ssh (needed for qemu+ssh libvirt connection).
+## Determine whether Ruby on Rails can execute ssh (needed for qemu+ssh libvirt connection).
 ## </p>
 ## </desc>
-gen_tunable(passenger_can_spawn_ssh, true)
+gen_tunable(foreman_rails_can_spawn_ssh, true)
 
 ## <desc>
 ## <p>
-## Determine whether passenger can connect to http_port_t when preloading app.
+## Determine whether Ruby on Rails can connect to http_port_t when preloading app.
 ## </p>
 ## </desc>
-gen_tunable(passenger_can_connect_http, true)
+gen_tunable(foreman_rails_can_connect_http, true)
 
 ## <desc>
 ## <p>
-## Determine whether passenger can connect to puppet_port_t when preloading app.
+## Determine whether Ruby on Rails can connect to local or remote libvirt.
 ## </p>
 ## </desc>
-gen_tunable(passenger_can_connect_puppetmaster, true)
+gen_tunable(foreman_rails_can_connect_libvirt, true)
 
 ## <desc>
 ## <p>
-## Determine whether passenger can connect to local or remote libvirt.
+## Determine whether Ruby on Rails can connect to remote SMTP services
 ## </p>
 ## </desc>
-gen_tunable(passenger_can_connect_libvirt, true)
+gen_tunable(foreman_rails_can_connect_smtp, true)
 
-## <desc>
-## <p>
-## Determine whether passenger can connect to remote SMTP services
-## </p>
-## </desc>
-gen_tunable(passenger_can_connect_smtp, true)
+######################################
+#
+# Temporary fixes for RHEL base policy
+#
 
-# Some basic aliases for different aspects of the filesystem to make things
-# more clear.
+# RHBZ#1527522: logrotate does not allow signals
+gen_require(`
+    type logrotate_t;
+')
+systemd_config_generic_services(logrotate_t)
+
+ ###############
+#
+# Types
+#
+
+type foreman_rails_t;
+type foreman_rails_exec_t;
+
+type foreman_rails_unit_file_t;
+systemd_unit_file(foreman_rails_unit_file_t)
+
 require{
     type etc_t;
 }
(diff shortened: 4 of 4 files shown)


ekohl:
I've marked it as a draft. Please use the Ready for review when, well, it's ready for review :)

README.md
```diff
@@ -19,7 +19,7 @@ Make sure you provide the correct distribution. Possible values are:
 
 There's a target to do compile and load policy on a remote system via ssh:
 
-    make remote-load host=my.host.lan distro=rhel7 name=foreman
+    make remote-load HOST=sat68.nat.lan DISTRO=rhel7 NAME=foreman
```
ekohl: Nit: it'd be best to stay generic:
```suggestion
    make remote-load HOST=myhost.example.com DISTRO=rhel7 NAME=foreman
```

ekohl:
Note that with Foreman 2.1 we still support using Passenger to run Foreman itself. We're not at the point where we want to fully drop it because we don't know how it will run in practice. For Passenger we have a lot of tuning knowledge and guides. https://projects.theforeman.org/issues/29780 will track removal.

lzap:
> Note that with Foreman 2.1 we still support using Passenger to run Foreman itself.

Damn, I had no idea. I already started dropping Passenger rules :-(

So is the plan to support both for some time? @ehelms @ekohl Then I need to start over.

README.md
lzap: Fixed.

ekohl:
> So is the plan to support both for some time? @ehelms @ekohl Then I need to start over.

At this point we don't know how stable Puma will be in production. We've done some tests and it appears stable, but we still want a fallback plan in case it doesn't. Best case scenario it works great in 2.1 and we can drop it in 2.2. Worst case we fully need to go back to the drawing board. I've been running it on my personal server for a while so I'm quite confident it won't be worst case, but my personal server isn't really a huge scale.

lzap:
I have created a packaging patch for sidekiq: https://github.com/theforeman/foreman-packaging/pull/5255 make sure you have it when testing these policy changes.

lzap:
This is now ready for review, but two patches are needed - wrappers for both rails and sidekiq needed to test this.

lzap:
How to test:

* Merge Core PR (or apply patch locally): https://github.com/theforeman/foreman/pull/7666
* Build foreman RPMs and install it: https://github.com/theforeman/foreman-packaging/pull/5255 (depends on the above unfortunately)
* Checkout this branch here and from your laptop perform:

    make remote-load HOST=sat68.nat.lan DISTRO=rhel7 NAME=foreman

On the host (sat68.nat.lan) then:

    restorecon -RvF /
    setenforce 0
    foreman-maintain service restart

Both Puma and Sidekiq should run with this domain:

```
# ps auxfZ | grep foreman_rails_t
unconfined_u:unconfined_r:unconfined_t:s0-s0:c0.c1023 root 26393 0.0  0.0 112812 972 pts/0 S+ 11:21   0:00          \_ grep --color=auto foreman_rails_t
system_u:system_r:foreman_rails_t:s0 foreman 1243 0.7  0.7 1015344 101456 ?    Ssl  09:40   0:47 puma 4.3.3 (tcp://127.0.0.1:3000) [foreman]
system_u:system_r:foreman_rails_t:s0 foreman 3234 0.2  5.0 2032400 649744 ?    Sl   09:44   0:16  \_ puma: cluster worker 0: 1243 [foreman]
system_u:system_r:foreman_rails_t:s0 foreman 26193 0.0  1.4 1754868 187952 ?   Sl   11:09   0:00  \_ puma: cluster worker 1: 1243 [foreman]
system_u:system_r:foreman_rails_t:s0 foreman 24673 0.0  0.0 4356  316 ?        Ss   10:35   0:00 /usr/bin/scl enable tfm -- sidekiq -e production -r /usr/share/foreman/extras/dynflow-sidekiq.rb -C /etc/foreman/dynflow/orchestrator.yml
system_u:system_r:foreman_rails_t:s0 foreman 24686 0.0  0.0 113284 548 ?       S    10:35   0:00  \_ /bin/bash /var/tmp/scl5TR0WH
system_u:system_r:foreman_rails_t:s0 foreman 24763 1.8  4.0 966876 525468 ?    Sl   10:35   0:49      \_ sidekiq 5.2.7  [0 of 1 busy]
system_u:system_r:foreman_rails_t:s0 foreman 25253 0.0  0.0 4356  316 ?        Ss   10:36   0:00 /usr/bin/scl enable tfm -- sidekiq -e production -r /usr/share/foreman/extras/dynflow-sidekiq.rb -C /etc/foreman/dynflow/worker.yml
system_u:system_r:foreman_rails_t:s0 foreman 25255 0.0  0.0 113284 552 ?       S    10:36   0:00  \_ /bin/bash /var/tmp/sclk61anm
system_u:system_r:foreman_rails_t:s0 foreman 25285 11.9  5.5 1179468 708672 ?  Sl   10:36   5:20      \_ sidekiq 5.2.7  [1 of 5 busy]
system_u:system_r:foreman_rails_t:s0 foreman 25270 0.0  0.0 4356  312 ?        Ss   10:36   0:00 /usr/bin/scl enable tfm -- sidekiq -e production -r /usr/share/foreman/extras/dynflow-sidekiq.rb -C /etc/foreman/dynflow/worker-hosts-queue.yml
system_u:system_r:foreman_rails_t:s0 foreman 25273 0.0  0.0 113284 556 ?       S    10:36   0:00  \_ /bin/bash /var/tmp/scll5elyt
system_u:system_r:foreman_rails_t:s0 foreman 25303 1.7  4.0 973764 517172 ?    Sl   10:36   0:46      \_ sidekiq 5.2.7  [0 of 5 busy]
```

Do some API/UI actions, there should be no denials except this one which was already reported for katello-selinux and fixed in nightly:

```
#!!!! This avc can be allowed using the boolean 'foreman_rails_can_connect_all'
allow foreman_rails_t candlepin_activemq_port_t:tcp_socket name_connect;
```

lzap:
Rebased file contexts, added libexec directory for the Rails wrapper.

foreman.fc
```diff
@@ -16,21 +16,27 @@
...
 
-/usr/share/foreman/.ssh(/.*)?           gen_context(system_u:object_r:ssh_home_t,s0)
-/usr/share/foreman/extras/noVNC/websockify\.py gen_context(system_u:object_r:websockify_exec_t,s0)
+/var/run/foreman(/.*)?                              gen_context(system_u:object_r:foreman_var_run_t,s0)
+
+/usr/share/foreman/.ssh(/.*)?                       gen_context(system_u:object_r:ssh_home_t,s0)
+/usr/share/foreman/extras/noVNC/websockify\.py      gen_context(system_u:object_r:websockify_exec_t,s0)
+
+/usr/share/foreman/bin/(rails|sidekiq)          --  gen_context(system_u:object_r:foreman_rails_exec_t,s0)
+/usr/libexec/foreman/bin/tfm-rails              --  gen_context(system_u:object_r:foreman_rails_exec_t,s0)
```
adamruzicka: According to https://github.com/theforeman/foreman-packaging/pull/5255/files#diff-3399e5c03dc282964481f8f3ced9c141R772 this should be `/usr/libexec/foreman/tfm-rails`

lzap: Right I am testing this by creating those files manually and I used wrong path. Let me start over creating all of them.

lzap:
There was one more problem, in order to avoid spawning `/bin/bash` I had to modify the sed in SPEC file. Instead of sed new file is written completely:

```
  cat > extras/sidekiq <<-EOF
  #!/bin/bash
  source scl_source enable %{scl_name}
  exec sidekiq "\$@"
EOF
```

lzap:
I am running into issues with https://serverfault.com/questions/960132/puma-with-systemd-and-socket-activation-errnoeaddrinuse

ekohl:
My experience tells me that happens when the file descriptor isn't passed on. Your wrapper is certainly the cause of this because I saw the exact same thing when there was a wrapper in place before.

lzap:
@adamruzicka @ekohl let me comment here but this is comment for this PR and the packaging/core PRs associated with this.

After long experimenting, I've came to this solution which is IMHO the best and cleanest.

A script `/usr/libexec/foreman/tfm-ruby-selinux` will be created by the SPEC file. It is simply a copy of `tfm-ruby` which is shipped with our SCL, but it has the required SELinux label on it. We canont use symlinks, SELinux does not allow that:

```
[root@sat68 ~]# ll -Z /usr/libexec/foreman/tfm-ruby-selinux 
-rwxr-xr-x. root root unconfined_u:object_r:foreman_rails_exec_t:s0 /usr/libexec/foreman/tfm-ruby-selinux

[root@sat68 ~]# cat /usr/libexec/foreman/tfm-ruby-selinux
#!/bin/bash
source scl_source enable tfm
exec ruby "$@"
```

File `/usr/share/foreman/bin/rails` will be modified by the SPEC so the shebang is set to `#!/usr/libexec/foreman/tfm-ruby-selinux`. This way the transition of `foreman.service` is nice and clean. No intermediate shells.

Note that `/usr/share/foreman/bin/rails` file must be in `/usr/share/foreman` because the script relies on the file directory, it cannot be moved. For sidekiq the situation is different - the startup will work outside of the Rails directory. So although @ekohl wanted me to create a helper script in `/usr/share`, I think it's better to put it where it belongs from the day one.

For sidekiq, new helper file `/usr/libexec/foreman/tfm-sidekiq-selinux` will be created by the SPEC in the very same fashion, except it calls sidekiq instead of ruby. And service file will be modifed by the SPEC as we do today to use this helper instead of "scl enable" which creates a temporary file.

```
[root@sat68 ~]# cat /usr/libexec/foreman/tfm-sidekiq-selinux 
#!/bin/bash
source scl_source enable tfm
exec sidekiq "$@"
```

This way all processes are in the target domain `foreman_rails_t` and also there are no ugly intermediate shell scripts which is particularly useful for sidekiq processes as there are many of them.

```
[root@sat68 ~]# ps auxfZ | grep foreman_rails_t
system_u:system_r:foreman_rails_t:s0 foreman 8709 7.3  3.9 1022484 510892 ?    Ssl  12:36   0:45 puma 4.3.3 (tcp://127.0.0.1:3000) [foreman]
system_u:system_r:foreman_rails_t:s0 foreman 8751 0.0  3.9 1760316 506380 ?    Sl   12:36   0:00  \_ puma: cluster worker 0: 8709 [foreman]
system_u:system_r:foreman_rails_t:s0 foreman 8756 0.0  3.9 1760316 506360 ?    Sl   12:36   0:00  \_ puma: cluster worker 1: 8709 [foreman]
system_u:system_r:foreman_rails_t:s0 foreman 9442 44.2  4.0 981104 518688 ?    Ssl  12:44   0:44 sidekiq 5.2.7  [0 of 1 busy]
system_u:system_r:foreman_rails_t:s0 foreman 9469 46.3  3.9 976964 509516 ?    Ssl  12:44   0:45 sidekiq 5.2.7  [0 of 5 busy]
system_u:system_r:foreman_rails_t:s0 foreman 9473 46.0  3.9 973588 507236 ?    Ssl  12:44   0:45 sidekiq 5.2.7  [0 of 5 busy]

[root@sat68 ~]# systemctl status dynflow-sidekiq@orchestrator.service
● dynflow-sidekiq@orchestrator.service - Foreman jobs daemon - orchestrator on sidekiq
   Loaded: loaded (/usr/lib/systemd/system/dynflow-sidekiq@.service; enabled; vendor preset: disabled)
   Active: active (running) since Thu 2020-05-21 12:40:53 CEST; 2min 0s ago
     Docs: https://theforeman.org
 Main PID: 8983 (sidekiq)
   CGroup: /system.slice/system-dynflow\x2dsidekiq.slice/dynflow-sidekiq@orchestrator.service
           └─8983 sidekiq 5.2.7  [0 of 1 busy]
```

Compare:

```
[root@sat68 ~]# systemctl status dynflow-sidekiq@orchestrator.service
● dynflow-sidekiq@orchestrator.service - Foreman jobs daemon - orchestrator on sidekiq
   Loaded: loaded (/usr/lib/systemd/system/dynflow-sidekiq@.service; enabled; vendor preset: disabled)
   Active: active (running) since Thu 2020-05-21 12:30:28 CEST; 5min ago
     Docs: https://theforeman.org
 Main PID: 8571 (scl)
   CGroup: /system.slice/system-dynflow\x2dsidekiq.slice/dynflow-sidekiq@orchestrator.service
           ├─8571 /usr/bin/scl enable tfm sidekiq -e production -r /usr/share/foreman/extras/dynflow-sidekiq.rb -C /etc/foreman/dynflow/orchestrator.yml
           ├─8572 /bin/bash /var/tmp/scldwnutq
           └─8586 sidekiq 5.2.7  [0 of 1 busy]
```

I am going to do the changes in packaging and this PR. Note I will close the core PR.

lzap:
Pushed here and into the packaging branch as well, this is now ready for re-review and re-testing @adamruzicka

foreman.fc
```diff
@@ -16,21 +16,27 @@
...
 #
 
-# Foreman file contexts
+# Foreman and Ruby on Rails
 
-/etc/foreman(/.*)?                      gen_context(system_u:object_r:foreman_config_t,s0)
-/etc/puppet/node.rb                 --  gen_context(system_u:object_r:foreman_enc_t, s0)
-/etc/puppetlabs/puppet/node.rb      --  gen_context(system_u:object_r:foreman_enc_t, s0)
+/usr/lib/systemd/system/foreman.*               --  gen_context(system_u:object_r:foreman_rails_unit_file_t,s0)
+/etc/systemd/system/foreman.*                   --  gen_context(system_u:object_r:foreman_rails_unit_file_t,s0)
```
adamruzicka, line 22: Why do we need our own type for service unit files instead of `systemd_unit_file_t`?

adamruzicka:
With somes changes to the deployed files described in the packaging PR and masking `foreman.socket` this works nicely.

adamruzicka:
Looks like I was too hasty. I'm getting this when refreshing features on a smart proxy

```
ERF50-4354 [Foreman::WrappedException]: Unable to read SSL certification or key for proxy communication, check settings for ssl_certificate, ssl_ca_file and ssl_priv_key and ensure they are readable by the foreman user. ([Errno::EACCES]: Permission denied @ rb_sysopen - /etc/puppetlabs/puppet/ssl/certs/centos-selinux-2.example.com.pem)
/usr/share/foreman/lib/proxy_api/resource.rb:139:in `rescue in ssl_auth_params'
/usr/share/foreman/lib/proxy_api/resource.rb:125:in `ssl_auth_params'
/usr/share/foreman/lib/proxy_api/resource.rb:23:in `initialize'
/opt/theforeman/tfm/root/usr/share/gems/gems/foreman_remote_execution-3.2.1/app/lib/proxy_api/remote_execution_ssh.rb:5:in `initialize'
/opt/theforeman/tfm/root/usr/share/gems/gems/foreman_remote_execution-3.2.1/app/models/concerns/foreman_remote_execution/smart_proxy_extensions.rb:10:in `new'
/opt/theforeman/tfm/root/usr/share/gems/gems/foreman_remote_execution-3.2.1/app/models/concerns/foreman_remote_execution/smart_proxy_extensions.rb:10:in `update_pubkey'
/opt/theforeman/tfm/root/usr/share/gems/gems/foreman_remote_execution-3.2.1/app/models/concerns/foreman_remote_execution/smart_proxy_extensions.rb:21:in `refresh'
/usr/share/foreman/app/controllers/smart_proxies_controller.rb:34:in `refresh'
```

Looks like foreman can't read its certs
```
[root@centos-selinux-2 ~]# ls -laZ /etc/puppetlabs/puppet/ssl/certs/centos-selinux-2.example.com.pem
-rw-r--r--. puppet puppet system_u:object_r:puppet_etc_t:s0 /etc/puppetlabs/puppet/ssl/certs/centos-selinux-2.example.com.pem
```
audit.log
```
type=AVC msg=audit(1590142276.268:314): avc:  denied  { search } for  pid=3498 comm="diagnostic_con*" name="puppetlabs" dev="vda3" ino=221443 scontext=system_u:system_r:foreman_rails_t:s0 tcontext=system_u:object_r:puppet_etc_t:s0 tclass=dir permissive=0
type=SYSCALL msg=audit(1590142276.268:314): arch=c000003e syscall=2 success=no exit=-13 a0=7f15cc1e8160 a1=80000 a2=1b6 a3=d items=0 ppid=3412 pid=3498 auid=4294967295 uid=996 gid=993 euid=996 suid=996 fsuid=996 egid=993 sgid=993 fsgid=993 tty=(none) ses=4294967295 comm="diagnostic_con*" exe="/opt/rh/rh-ruby25/root/usr/bin/ruby" subj=system_u:system_r:foreman_rails_t:s0 key=(null)
```

lzap:
[](url)

> masking `foreman.socket` this works nicely.

Wait, why woudl you mask socket? We want socket activation, it works for me.

Nice catch, this makes sense the rule `puppet_read_config(foreman_rails_t)` was removed as I thought this was puppetmaster only.

Fixed, pushed. You can actually use permissive, the policy should be stable enough just check denials. FYI this is how you can check missing rules:

```
yum -y install setools-console policycoreutils-python policycoreutils selinux-policy-devel
audit2allow -Rla

require {
        type foreman_rails_t;
        type candlepin_activemq_port_t;
        class tcp_socket name_connect;
}

#============= foreman_rails_t ==============

#!!!! This avc can be allowed using the boolean 'foreman_rails_can_connect_all'
allow foreman_rails_t candlepin_activemq_port_t:tcp_socket name_connect;
corenet_udp_bind_generic_node(foreman_rails_t)
```

This shows you missing rules from the last policy reload. This one is expected (known bug for katello-selinux).

adamruzicka:
Huh, not sure what I did wrong with the socket activation, perhaps I forgot to restart something, this showed up in my logs https://p.teknik.io/yIbw3 . Now I unmasked it, restarted the machine just to be sure and it works.

I'll keep playing with this a bit more, but so far so good.

lzap:
> `for_fd': Bad file descriptor - not a socket file descriptor (Errno::EBADF)

You can read more about this error in the links I've shown above, but in short - one of my versions used intermediate shell script which throws away a socket which is passed to the subprocess and since in our configuration we dont use named UNIX or TCP socket for IPC, it does not work.

foreman-selinux-relabel
```diff
@@ -11,8 +11,11 @@
   /etc/puppetlabs/puppet/node.rb \
   /etc/sysconfig/foreman* \
   /etc/rc.d/init.d/foreman* \
+  /usr/lib/systemd/system/foreman* \
+  /etc/systemd/system/foreman* \
   /etc/logrotate.d/foreman* \
   /etc/cron.d/foreman* \
+  /usr/libexec/foreman \
   /usr/lib{64,}/ruby/gems/1.8/gems/passenger-* \
```
ekohl, line 18: This can be dropped since it's EL6 code. Same for all other Ruby 1.8 stuff. Feel free to do that in a later PR.

foreman-selinux-relabel
```diff
@@ -11,8 +11,11 @@
   /etc/puppetlabs/puppet/node.rb \
   /etc/sysconfig/foreman* \
   /etc/rc.d/init.d/foreman* \
```
ekohl: We don't ship this anymore

foreman.te
```diff
@@ -71,55 +61,82 @@ gen_tunable(websockify_can_connect_all, false)
...
+gen_tunable(foreman_rails_can_connect_smtp, true)
 
 ## <desc>
 ## <p>
-## Determine whether passenger can connect to remote SMTP services
+## Deprecated SQLite3 support
 ## </p>
 ## </desc>
-gen_tunable(passenger_can_connect_smtp, true)
+gen_tunable(foreman_rails_can_manage_sqlite, false)
```
ekohl: We've dropped foreman-sqlite3 in 2.1 so you don't need this unless you want to cherry pick this to 2.0.

foreman.te
```diff
@@ -171,190 +188,156 @@ require{
...
-# or to connect to remote Cockpit instance (via Remote Execution)
-allow passenger_t websm_port_t:tcp_socket name_connect;
+# It is not possible to do transition into SCL application
+# without calling a shell: https://github.com/sclorg/scl-utils/issues/31
+# These rules are for: /usr/libexec/foreman/tfm-ruby-selinux
+corecmd_exec_shell(foreman_rails_t)
+corecmd_exec_ls(foreman_rails_t)
+# The following is needed for "scl enable" which creates temporary
+# shell script /var/tmp/xxxxxxx and a subprocess.
+files_manage_generic_tmp_files(foreman_rails_t)
```
ekohl: Is this still the case with our `scl_source` + exec wrappers? We dropped `/usr/libexec/foreman/tfm-ruby-selinux`

foreman.te
```diff
@@ -171,190 +188,156 @@ require{
...
+logging_create_devlog_dev(foreman_rails_t)
+
+# Systemd socket activation (Puma called with LISTEN_FDS environmental variable)
+allow foreman_rails_t self:netlink_route_socket { all_netlink_route_socket_perms };
+allow foreman_rails_t self:tcp_socket { all_tcp_socket_perms };
+allow foreman_rails_t self:udp_socket { all_udp_socket_perms };
+allow foreman_rails_t self:unix_dgram_socket { all_unix_dgram_socket_perms };
+corenet_tcp_bind_generic_node(foreman_rails_t)
+
+# Listening for HTTP communication on port 3000 (development default port)
```
ekohl: Also the one we currently use in production

foreman.te
```diff
@@ -171,190 +188,156 @@ require{
...
-        write_files_pattern(passenger_t, foreman_db_t, foreman_db_t)
+# rubygem FFI is used in foreman-tasks (BZ#1670109 / RM#26951)
+allow foreman_rails_t self:process execmem;
 
-        # Allow Foreman to do DNS queries via Ruby stdlib net package
-        # http://projects.theforeman.org/issues/8030
-        corenet_udp_bind_generic_port(passenger_t)
+# Connecting to Foreman Proxy on port 9090 (Katello)
+# or to remote Cockpit instance (via Remote Execution)
+allow foreman_rails_t websm_port_t:tcp_socket name_connect;
```
ekohl: Doesn't this belong in `katello-selinux`?

foreman.te
```diff
@@ -171,190 +188,156 @@ require{
...
 ')
 
-tunable_policy(`httpd_run_foreman', `
-    admin_pattern(httpd_t, passenger_tmp_t, passenger_tmp_t)
-    admin_pattern(passenger_t, httpd_tmp_t, httpd_tmp_t)
-')
+# Connecting to Redis
+corenet_tcp_connect_redis_port(foreman_rails_t)
 
+# Connecting to puppet server
```
ekohl: Foreman never connects to the puppetmaster. Only foreman-proxy does

foreman.te
```diff
@@ -171,190 +188,156 @@ require{
...
-    # Transition to puppet_master
-    corecmd_search_bin(passenger_t)
-    domtrans_pattern(passenger_t, puppetmaster_exec_t, puppetmaster_t)
-')
+#################
+# TODO
+#
+# system_u:system_r:unconfined_service_t:s0 foreman 1227 0.0  0.0 4356 584 ?     Ss   10:26   0:00 /usr/bin/scl enable tfm sidekiq -e production -r /usr/share/foreman/extras/dynflow-sidekiq.rb 
+# system_u:system_r:unconfined_service_t:s0 foreman 1230 0.0  0.0 113284 1472 ?  S    10:26   0:00  \_ /bin/bash /var/tmp/sclPZHShi
+# system_u:system_r:unconfined_service_t:s0 foreman 1671 0.3  4.2 972624 545000 ? Sl  10:26   1:18      \_ sidekiq 5.2.7  [0 of 5 busy]
```
ekohl: Now that we have the wrapper, this is no longer an issue, right?

foreman.te
```diff
@@ -463,35 +450,28 @@ tunable_policy(`websockify_can_connect_all',`
 # Container / Docker
```
ekohl, line 414: We've dropped foreman_docker so is this still relevant?

foreman.te
```diff
@@ -501,34 +481,183 @@ typealias bin_t alias foreman_hook_t;
...
+    tunable_policy(`passenger_run_foreman', `
+        mta_send_mail(passenger_t)
+    ')
+')
+optional_policy(`
+    tunable_policy(`passenger_run_foreman', `
+        mysql_stream_connect(passenger_t)
+        mysql_list_db(passenger_t)
+    ')
+')
```
ekohl: MySQL was dropped in 2.0 so no need for this

foreman.te
```diff
@@ -501,34 +481,183 @@ typealias bin_t alias foreman_hook_t;
...
+    tunable_policy(`passenger_run_foreman', `
+        hostname_exec(passenger_t)
+    ')
+')
+optional_policy(`
+    tunable_policy(`passenger_run_foreman', `
+        corecmd_search_bin(passenger_t)
+        domtrans_pattern(passenger_t, puppetmaster_exec_t, puppetmaster_t)
+    ')
+')
```
ekohl, line 560: Foreman never talks to the Puppetmaster

ekohl (commented):
Small administrative note: the PR/commit title still says "drop passenger" which is no longer true.

There are various comments inline to which I don't know the exact answer but that jumped out to me.

adamruzicka:
This works nicely with the packaging PR that was just merged

foreman-selinux-relabel
lzap: Yeah I know the policy is in bad shape, I wanted to to "no changes commit" first since we are removing it pretty soon. I can do a followup commit where I simply remove it all and we would only cherry pick the first one into 2.1. Lemme see your other comments first.

foreman.te
lzap: Yeah removing.

foreman.te
lzap: Unfortunately yes, I was hoping that this will be not necessary but I still see a denial. Honestly, don't know which component does that, some subprocess? I stopped digging when I learned that we cannot live with the other two rules above which are far more dangerous than tmp files read write. So I suggest to live with this.

foreman.te
lzap: Changed the comment.

foreman.te
lzap: The comment maybe, but the rule does not - 9090 is a cockpit port. Let me swap those comments.

foreman.te
lzap: That was probably puppet master: https://projects.theforeman.org/issues/21887

It was passenger's infra preloading the app. Removed.

foreman.te
lzap: LOL I left these notes here, removed :-)

foreman.te
lzap: Removed. If you can confirm nothing in Katello connects to docker ports just to be sure.

lzap: Wait! For technical reason (upgrade path vs core policy) I will do this separately. There was some much pain made when we were introducing this redefined port, these are darkest things in SELinux and I would rather keep this removal separate because I am pretty sure removal will be painful too.

lzap: https://projects.theforeman.org/issues/29957

foreman.te
lzap: Ok it looks like you did not have too many comments, I will remove everything in the same commit then. Removed.

foreman.te
lzap: Well, we do carry whole policy for puppetmaster as it also runs in passenger (the old version). I am keeping these until passenger is completely removed. Granted probably all of our users uses the Java passenger.

This rulle is a domain transition for puppet master (the old version). If I'd remove this, I'd need to remove pretty much all of the legacy passenger stuff too, which is planned in a separate PR.

foreman-selinux-relabel
lzap: I don't know if you mean just this line or all of the lib64 ones, I'd keep it until we remove all of the passenger code in a separate PR if you don't mind.

foreman-selinux-relabel
lzap: Removed.

lzap:
Okay, rebased these are minor changes so it should work. Ready to go! I will backport this into 2.1 once it's merged.

foreman-selinux-relabel
ekohl: The entire line can go. `/usr/lib{64,}/ruby/gems/1.8` existed on EL6 which had Ruby 1.8.7 but luckily we're beyond those dark days.

foreman.fc
```diff
@@ -16,21 +16,27 @@
...
-/var/log/foreman(/.*)?                  gen_context(system_u:object_r:foreman_log_t,s0)
+/var/lib/foreman(/.*)?                              gen_context(system_u:object_r:foreman_lib_t,s0)
+/var/lib/foreman/db/(.*.sqlite3)?                   gen_context(system_u:object_r:foreman_db_t,s0)
 
-/var/run/foreman(/.*)?                  gen_context(system_u:object_r:foreman_var_run_t,s0)
+/var/log/foreman(/.*)?                              gen_context(system_u:object_r:foreman_log_t,s0)
 
-/usr/share/foreman/.ssh(/.*)?           gen_context(system_u:object_r:ssh_home_t,s0)
-/usr/share/foreman/extras/noVNC/websockify\.py gen_context(system_u:object_r:websockify_exec_t,s0)
+/var/run/foreman(/.*)?                              gen_context(system_u:object_r:foreman_var_run_t,s0)
```
ekohl, line 32: General question: `/var/run` is a symlink to `/run` nowadays. Does this still make sense? I don't mind keeping it in since packaging still uses that as well, even though it doesn't make sense anymore.

foreman.fc
```diff
@@ -16,21 +16,27 @@
...
 
-/var/lib/foreman(/.*)?                  gen_context(system_u:object_r:foreman_lib_t,s0)
-/var/lib/foreman/db/(.*.sqlite3)?       gen_context(system_u:object_r:foreman_db_t,s0)
+/etc/foreman(/.*)?                                  gen_context(system_u:object_r:foreman_config_t,s0)
+/etc/puppet/node.rb                             --  gen_context(system_u:object_r:foreman_enc_t, s0)
+/etc/puppetlabs/puppet/node.rb                  --  gen_context(system_u:object_r:foreman_enc_t, s0)
 
-/var/log/foreman(/.*)?                  gen_context(system_u:object_r:foreman_log_t,s0)
+/var/lib/foreman(/.*)?                              gen_context(system_u:object_r:foreman_lib_t,s0)
+/var/lib/foreman/db/(.*.sqlite3)?                   gen_context(system_u:object_r:foreman_db_t,s0)
```
ekohl: Since you dropped the other sqlite code, this can also go if you want to clean it up completely.

foreman.te
ekohl: I'm guessing `scl_source` generates that then.

Should we hide that behind an EL7 conditional since EL8 doesn't use SCLs?

foreman.te
```diff
@@ -171,189 +174,134 @@ require{
...
+# Certificates (Foreman uses puppet certificates as well)
+miscfiles_read_certs(foreman_rails_t)
+puppet_read_config(foreman_rails_t)
 
-# Allow Foreman to connect to Foreman Proxy on port 9090 (Katello)
-# or to connect to remote Cockpit instance (via Remote Execution)
-allow passenger_t websm_port_t:tcp_socket name_connect;
+# It is not possible to do transition into SCL application
+# without calling a shell: https://github.com/sclorg/scl-utils/issues/31
+# These rules are for: /usr/libexec/foreman/tfm-ruby-selinux
```
ekohl: This particular file doesn't exist so that's a bit confusing.

foreman.te
```diff
@@ -171,189 +174,134 @@ require{
...
-dev_read_rand(passenger_t)
+# Connecting to hosts and guests
+corenet_tcp_connect_virt_port(foreman_rails_t)
+corenet_tcp_connect_ssh_port(foreman_rails_t)
 
-# Passenger 6 is setting socket context explicitly in
-# https://github.com/phusion/passenger/blob/master/src/agent/Core/CoreMain.cpp#L376-L406
-allow passenger_t self:process setsockcreate;
+# Allow Foreman to do DNS queries via Ruby stdlib net package
+# http://projects.theforeman.org/issues/8030
```
ekohl:
```suggestion
# https://projects.theforeman.org/issues/8030
```

ekohl (commented):
Overall I think this is pretty ok. It may allow more than strictly needed, but compared to running unconfined that doesn't matter too much.

foreman.fc
lzap: Honestly, I don't know I'd rather avoid changes that late to the release as we need to backport this one.

foreman.fc
lzap: Hmmm right but one thing comes to my mind. We do actually use sqlite3 during build process, but if my understanding is correct the process is not yet confined therefore even if SELinux is turned on on the build machine it should work fine. Removed.

foreman.te
lzap: A leftover, removed.

foreman.te
lzap: Done.

lzap:
Rebased only cosmetic changes, no need to retest this.

```diff
diff --git a/foreman.fc b/foreman.fc
index 1d663eb..3ff8a6a 100644
--- a/foreman.fc
+++ b/foreman.fc
@@ -26,7 +26,6 @@
 /etc/puppetlabs/puppet/node.rb                  --  gen_context(system_u:object_r:foreman_enc_t, s0)
 
 /var/lib/foreman(/.*)?                              gen_context(system_u:object_r:foreman_lib_t,s0)
-/var/lib/foreman/db/(.*.sqlite3)?                   gen_context(system_u:object_r:foreman_db_t,s0)
 
 /var/log/foreman(/.*)?                              gen_context(system_u:object_r:foreman_log_t,s0)
 
diff --git a/foreman.te b/foreman.te
index ff73205..c4a690d 100644
--- a/foreman.te
+++ b/foreman.te
@@ -198,7 +198,6 @@ puppet_read_config(foreman_rails_t)
 
 # It is not possible to do transition into SCL application
 # without calling a shell: https://github.com/sclorg/scl-utils/issues/31
-# These rules are for: /usr/libexec/foreman/tfm-ruby-selinux
 corecmd_exec_shell(foreman_rails_t)
 corecmd_exec_ls(foreman_rails_t)
 # The following is needed for "scl enable" which creates temporary
@@ -260,7 +259,7 @@ corenet_tcp_connect_virt_port(foreman_rails_t)
 corenet_tcp_connect_ssh_port(foreman_rails_t)
 
 # Allow Foreman to do DNS queries via Ruby stdlib net package
-# http://projects.theforeman.org/issues/8030
+# https://projects.theforeman.org/issues/8030
 corenet_udp_bind_generic_port(foreman_rails_t)
 
 # rubygem FFI is used in foreman-tasks (BZ#1670109 / RM#26951)
```

adamruzicka (approved):
Looks good to me, let's get this in

adamruzicka:
Thank you @lzap & @ekohl !

lzap:
Thanks for testing! :100:

lzap:
CP 2.1: https://github.com/theforeman/foreman-selinux/pull/102

foreman.fc
ekohl: > We do actually use sqlite3 during build process

Nope, that was changed. Since 2.1 we use nulldb during builds

PR #107, opened 2020-07-01
Fixes #30279 - allow http(s) proxy ports by default

lzap:
Previously this was allowed in the base policy for passenger, now since we moved to puma these rules were actually missing.

### DIFF
--- a/foreman.te
+++ b/foreman.te
@@ -89,6 +89,13 @@ gen_tunable(foreman_rails_can_connect_http, true)
 
 ## <desc>
 ## <p>
+## Determine whether Ruby on Rails can connect to http_cache_port_t and squid_port_t.
+## </p>
+## </desc>
+gen_tunable(foreman_rails_can_connect_http_proxy, true)
+
+## <desc>
+## <p>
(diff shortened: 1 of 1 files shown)


adamruzicka (approved):
Makes sense, ACK

adamruzicka:
Thank you @lzap!

ekohl:
Why is this needed? Is this for when you have configured an outgoing HTTP proxy in the Foreman application? If so, it probably should be cherry picked into 2.1. AFAIK that was already tagged so that would be a known issue in 2.1.0.

cc @tbrisker

ehelms:
I was also curious where this applied. Is this for running an HTTP proxy on the same box as your Foreman application?

lzap:
@ekohl yes, when foreman is configured to connect to a HTTP(s) proxy, we should cherry pick this, yeah. There is a BZ associated with this. It looks like a regression.

@ehelms not sure what you mean, but this is a rule for *outgoing* connection

tbrisker:
2.1 - 073a56d

lzap:
Thanks, QA ran into two more issues :facepalm:  One should be already fixed (Katello) and the other is new.

PR #108, opened 2020-07-13
Fixes #30378 - added uploader.sh typealias

lzap:
Please follow the README to test this (the `make` command).

### DIFF
--- a/foreman.fc
+++ b/foreman.fc
@@ -69,3 +69,6 @@
 
 /usr/share/gems/gems/foreman-tasks-.*/bin/foreman-tasks -- gen_context(system_u:object_r:foreman_tasks_exec_t,s0)
 /opt/theforeman/tfm/root/usr/share/gems/gems/foreman-tasks-.*/bin/foreman-tasks -- gen_context(system_u:object_r:foreman_tasks_exec_t,s0)
+
+# Red Hat Cloud plugin
+/var/lib/foreman/red_hat_inventory/uploads/uploader.sh -- gen_context(system_u:object_r:foreman_uploader_bin_t,s0)
--- a/foreman.te
+++ b/foreman.te
@@ -483,6 +483,13 @@ typealias bin_t alias foreman_hook_t;
 
 ######################################
 #
+# Foreman Red Hat Cloud plugin
+#
+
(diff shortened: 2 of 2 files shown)


foreman.fc
```diff
@@ -69,3 +69,6 @@
 
 /usr/share/gems/gems/foreman-tasks-.*/bin/foreman-tasks -- gen_context(system_u:object_r:foreman_tasks_exec_t,s0)
 /opt/theforeman/tfm/root/usr/share/gems/gems/foreman-tasks-.*/bin/foreman-tasks -- gen_context(system_u:object_r:foreman_tasks_exec_t,s0)
+
+# Red Hat Cloud plugin
+/var/lib/foreman/red_hat_inventory/uploads/uploader.sh -- gen_context(system_u:object_r:foreman_uploader_bin_t,s0)
```
ekohl, line 74: Before we add this, I'd like to investigate alternatives. This morning with @ShimShtein and @adamruzicka we talked about this. Other complications that we may run into is that `/var/lib` might be mounted `noexec` in which case it'd also fail.

adamruzicka: What would be the best location for it? It needs to be a place where it makes sense (something like `/usr/libexec`) but at the same time is writable by the rails process.

ekohl: My first question is why it needs to be dynamically generated.

ekohl:
The RH Cloud Plugin depends on Katello. Should we keep it in the katello-selinux policy?

foreman.fc
ShimShtein: @ekohl, it's for disconnected systems. The idea is to generate a script that can be downloaded alongside with the payload and executed with minimal user interaction. For this reason I "hard code" some variables into the script, so the user will not need to input them.

lzap: I had the very same question, you can still parametrize a short shell script as needed.

I agree that proper solution is to make this a generic script in `/usr/lib/libexec`.

lzap:
> The RH Cloud Plugin depends on Katello. Should we keep it in the katello-selinux policy?

I can move that, had no idea.

lzap:
I moved this to katello-selinux: https://github.com/Katello/katello-selinux/pull/27

PR #112, opened 2020-08-21
Fixes #30675 - allow readonly RPM commands

lzap:
SSIA, safe to merge. Tested.

The kerberos rule kept appearing, I am not sure what causes it, it is probably during boot. But analysis of this macros shows it only allows reading of /etc/krb5.conf which is safe to do.


lzap:
@adamruzicka tested can you give it a quick merge? This is for 6.8. Thanks.

adamruzicka (approved):
Makes sense, ACK

adamruzicka:
Thank you @lzap!

tbrisker:
2.2 - c6994f7
2.1 - f697d3a

PR #114, opened 2020-09-15
Fixes #30845 - allow DNS over TCP

lzap:
When server exceeds UDP packet limit, the DNS client uses TCP instead. We only had basic macros, more generic sysnet_dns_name_resolve will do it (includes DNSSEC support too). @ekohl

### DIFF
--- a/foreman.te
+++ b/foreman.te
@@ -269,6 +269,10 @@ corenet_tcp_connect_ssh_port(foreman_rails_t)
 corenet_udp_bind_generic_port(foreman_rails_t)
 corenet_udp_bind_generic_node(foreman_rails_t)
 
+# Allow to perform TCP DNS search when response is too big (BZ#1877443)
+# The following macro incudes corenet_tcp_connect_dns_port(foreman_rails_t)
+sysnet_dns_name_resolve(foreman_rails_t)
+
 # rubygem FFI is used in foreman-tasks (BZ#1670109 / RM#26951)
 allow foreman_rails_t self:process execmem;
 


foreman.te
```diff
@@ -269,6 +269,10 @@ corenet_tcp_connect_ssh_port(foreman_rails_t)
 corenet_udp_bind_generic_port(foreman_rails_t)
 corenet_udp_bind_generic_node(foreman_rails_t)
```
ekohl, line 270: Are these still needed? The comment suggests this was only for DNS

ekohl (approved):
Allowing DNS over TCP should indeed be done. Small question about cleaning up other rules, but in general :+1:

foreman.te
lzap: Yes, I don't want to tackle them in this PR, I was quickly looking into this and probably this could be removed as a separate PR. But I fail to find those macros in sysnet_dns_name_resolve and I can't find any other suitable macro to replace these.

lzap: Created https://projects.theforeman.org/issues/30849

tbrisker:
The original reporter confirmed this indeed fixes the issue, merging.

jasondickerson:
This has in fact resolved the AD Auth issue with looking up the AD domain A record with 42 IP's.  Thanks guys!

tbrisker:
2.2 - 9a54376
2.1 - b76189e

lzap:
I am pretty sure you are trolling us with those 42 records but hey, I love it! :-) Thanks for testing.

jasondickerson:
> I am pretty sure you are trolling us with those 42 records but hey, I love it! :-) Thanks for testing.

LOL.  I'm actually not.  This is on a large AD Domain with 42 domain controllers.  Thus, the domain A record, as setup by AD returns 42 A record answers.  What is interesting is that if you specify a DNS host that only returns 1 A record it works.  This makes me wonder if it is the amount of DNS A records returned, or just if more than 1 are returned.  I have no way to test that in the environment.  But could be good food for thought.  Thanks lzap!

lzap:
>  What is interesting is that if you specify a DNS host that only returns 1 A record it works.

Read this PR description to understand the reason! Thanks again, cheers!

ekohl:
> This makes me wonder if it is the amount of DNS A records returned, or just if more than 1 are returned.

DNS defaults to UDP, but UDP has no fragmentation so if a response is larger than a single UDP packet, implementations switch over to TCP. That's why you see differences.

PR #115, opened 2020-09-18
Fixes #30866 - enable and backport passenger rules

lzap:
This patch prepares us for a smooth experience if we end up using passenger instead of puma for the next release or two:

* All patches since Puma switch have been backported also to Passenger optional section.
* Passenger rules are turned on by default, we can deprecate them later. No need of installer changes to enable this boolean.

### DIFF
--- a/foreman.te
+++ b/foreman.te
@@ -28,7 +28,7 @@ policy_module(foreman, @@VERSION@@)
 ## Allow Ruby on Rails to run foreman (deprecated).
 ## </p>
 ## </desc>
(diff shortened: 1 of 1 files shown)


lzap:
@ekohl @ares

ekohl (approved):

tbrisker:
2.2 - 1c277a5
2.1 - 462b0ed

lzap:
I am so slow keeping up with my GH inbox, thanks a bunch!

PR #116, opened 2020-10-13
Add compilation GH action

lzap:
This is an attempt to write a compilation GH action on ubuntu OS. Wish me luck.

### DIFF
--- a/.github/workflows/compile.yaml
+++ b/.github/workflows/compile.yaml
@@ -1,0 +1,17 @@
+---
+name: Compile policy
+
+on: [pull_request]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Install SELinux
+        run: sudo apt-get install policycoreutils selinux-utils selinux-basics
+
+      - name: Compile policy for RHEL7
+        run: make DISTO=rhel7


ekohl:
I'd suggest to copy https://github.com/pulp/pulpcore-selinux/commit/41d4bba87fcbb174f7c24dccce7e8891b5856d66 instead so you actually compile on EL7 and EL8.

lzap:
Nice! Will do a followup PR.

lzap:
https://github.com/theforeman/foreman-selinux/pull/117

.github/workflows/compile.yaml
```diff
@@ -0,0 +1,17 @@
...
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Install SELinux
+        run: sudo apt-get install policycoreutils selinux-utils selinux-basics
+
+      - name: Compile policy for RHEL7
+        run: make DISTO=rhel7
```
adamruzicka, line 17: Shouldn't it be `DISTRO`?

lzap: Yes fixed in the other PR.

lzap:
FYI I had to merge this otherwise GH Actions won't trigger. I could not tell if it was working at all.

ekohl:
> FYI I had to merge this otherwise GH Actions won't trigger. I could not tell if it was working at all.

What I usually do is to also add the `push` action and check how it runs on my own fork.

PR #117, opened 2020-10-13
Update compilation GH action

lzap:
Use containers to build the policy, which is closer to our environment. I had no indea this could be done, container in a container?

### DIFF
--- a/.github/workflows/compile.yaml
+++ b/.github/workflows/compile.yaml
@@ -1,17 +1,28 @@
 ---
 name: Compile policy
 
-on: [pull_request]
+on:
+  - pull_request
+  - push
 
 jobs:
   build:
     runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        centos: ["7", "8"]
+    container:
+      image: centos:${{ matrix.centos }}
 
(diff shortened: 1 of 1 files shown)


ekohl:
> I had no indea this could be done, container in a container?

My understanding is that it loads the specified container instead of some other container.

lzap:
> My understanding is that it loads the specified container instead of some other container.

Interesting, they only list Ubuntu, Windows and Mac in their docs. I will use CentOS from now on for all my tasks.

lzap:
Nice it works! Great. I've added check for shell syntax of all scripts. Mind taking a look @ekohl ?

ekohl:
> Interesting, they only list Ubuntu, Windows and Mac in their docs. I will use CentOS from now on for all my tasks.

Note you can't use other actions such as python or ruby versions with such a container. Since it's a container you also won't have the kernel.

ekohl (approved):

lzap:
> won't have the kernel

Well, you have one. Obviously. Running in a namespace tho. :-) Thanks.

ekohl:
Fair enough. Not the matching kernel ;)

PR #118, opened 2020-11-26
make /run/foreman.sock match ALL files, not only regular

evgeni:
a socket is not a regular file, and thus this rule did not match on
(at least) EL8.

### DIFF
--- a/foreman.fc
+++ b/foreman.fc
@@ -30,7 +30,7 @@
 /var/log/foreman(/.*)?                              gen_context(system_u:object_r:foreman_log_t,s0)
 
 /var/run/foreman(/.*)?                              gen_context(system_u:object_r:foreman_var_run_t,s0)
-/run/foreman\.sock                              --  gen_context(system_u:object_r:foreman_var_run_t,s0)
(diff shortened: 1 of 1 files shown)


evgeni:
(untested, I am lazy)

evgeni:
this is supposed to fix the issues I've been seeing on EL8 in https://github.com/theforeman/puppet-foreman/pull/883#issuecomment-734168339

foreman.fc
```diff
@@ -30,7 +30,7 @@
 /var/log/foreman(/.*)?                              gen_context(system_u:object_r:foreman_log_t,s0)
 
 /var/run/foreman(/.*)?                              gen_context(system_u:object_r:foreman_var_run_t,s0)
-/run/foreman\.sock                              --  gen_context(system_u:object_r:foreman_var_run_t,s0)
+/run/foreman\.sock                                  gen_context(system_u:object_r:foreman_var_run_t,s0)
```
evgeni, line 33: this also could be `-s`, but then I don't know why it works on EL7…

lzap: It does? Weird, anyway this is okay I have never seen `-s` in use to be honest.

lzap (approved):
Good find, this one had to be tough to spot!

PR #119, opened 2020-11-26
match /var/run, not /run

evgeni:
/run is an alias for /var/run, so creating rules for /run doesn't work:

    # semanage fcontext --add -t foreman_var_run_t /run/foreman.sock
    ValueError: File spec /run/foreman.sock conflicts with equivalency rule '/run /var/run'; Try adding '/var/run/foreman.sock' instead

### DIFF
--- a/foreman.fc
+++ b/foreman.fc
@@ -30,7 +30,7 @@
 /var/log/foreman(/.*)?                              gen_context(system_u:object_r:foreman_log_t,s0)
 
 /var/run/foreman(/.*)?                              gen_context(system_u:object_r:foreman_var_run_t,s0)
-/run/foreman\.sock                                  gen_context(system_u:object_r:foreman_var_run_t,s0)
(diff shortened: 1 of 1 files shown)


foreman.fc
```diff
@@ -30,7 +30,7 @@
 /var/log/foreman(/.*)?                              gen_context(system_u:object_r:foreman_log_t,s0)
 
 /var/run/foreman(/.*)?                              gen_context(system_u:object_r:foreman_var_run_t,s0)
-/run/foreman\.sock                                  gen_context(system_u:object_r:foreman_var_run_t,s0)
+/var/run/foreman\.sock                                  gen_context(system_u:object_r:foreman_var_run_t,s0)
```
lzap: Can you fix the identation? Then I am good thanks.

lzap (commented):
Thanks.

PR #121, opened 2021-02-18
allow foreman_rails_t to touch rpm_temp

evgeni:
this should silence bundler induced selinux denials

### DIFF
--- a/foreman.te
+++ b/foreman.te
@@ -353,6 +353,18 @@ systemd_config_generic_services(logrotate_t)
 manage_files_pattern(foreman_rails_t, system_cronjob_tmp_t, system_cronjob_tmp_t)
 manage_dirs_pattern(foreman_rails_t, system_cronjob_tmp_t, system_cronjob_tmp_t)
 
+######################################
+#
+# RPM temp workaround
+# https://community.theforeman.org/t/foreman-nightly-rpm-pipeline-946-failed/
+
+gen_require(`
+    type rpm_script_tmp_t;
+')
+
+manage_files_pattern(foreman_rails_t, rpm_script_tmp_t, rpm_script_tmp_t)
+manage_dirs_pattern(foreman_rails_t, rpm_script_tmp_t, rpm_script_tmp_t)
+
 #######################################
 #
 # Remote Execution


ekohl (commented):
Just a random thought: should we enhance the systemd unit file to use `PrivateTmp=true`? A clean `/tmp` may also be a workaround.

evgeni:
Yes, probably a good idea, but not sure what else might break with that, so maybe a long term thing?

evgeni:
Or to phrase it differently: the introduced selinux test blocks all stable releases, so we either need to disable the test or fix it in a safe manner. I'd consider this PR "safe", whereas using `PrivateTmp=true` might have sideeffects which I'd prefer to let live in nightly for a while before considering.

ekohl:
I had the same thoughts.

Alternatively, what are the chances of fixing the actual CVE? IMHO we're really masking a real security issue this way.

lzap (approved):
Looks good, we usually keep types above but since this is a workaround, let's unblock us.

lzap:
This needs a RM ticket tho.

foreman.te
```diff
@@ -353,6 +353,18 @@ systemd_config_generic_services(logrotate_t)
...
+######################################
+#
+# RPM temp workaround
+# https://community.theforeman.org/t/foreman-nightly-rpm-pipeline-946-failed/
+
+gen_require(`
+    type rpm_script_tmp_t;
+')
+
+manage_files_pattern(foreman_rails_t, rpm_script_tmp_t, rpm_script_tmp_t)
```
ekohl, line 365: https://selinuxproject.org/page/AVCRules suggests you can use `dontaudit` instead of `allow` in a rule. `manage_files_pattern` is defined as
```
define(`manage_files_pattern',`
        allow $1 $2:dir rw_dir_perms;
        allow $1 $3:file manage_file_perms;
')
```

So should we use:
```suggestion
dontaudit foreman_rails_t rpm_script_tmp_t:dir rw_dir_perms
dontaudit foreman_rails_t rpm_script_tmp_t:file manage_file_perms
```

foreman.te
```diff
@@ -353,6 +353,18 @@ systemd_config_generic_services(logrotate_t)
...
+manage_dirs_pattern(foreman_rails_t, rpm_script_tmp_t, rpm_script_tmp_t)
```
ekohl, line 366: Similar here `manage_dirs_pattern` is
```
define(`manage_dirs_pattern',`
        allow $1 $2:dir rw_dir_perms;
        allow $1 $3:dir manage_dir_perms;
')
```

ekohl (commented):
@lzap can you advise here?

ekohl:
I opened https://projects.theforeman.org/issues/31922 for this.

lzap:
I don't understand the suggestion. Allow and dontaudit are two completely separate things, are we talking about actually allowing or disallowing those rules?

evgeni:
#122 is the better fix and also has an explanation *why* we don't need to allow those, just hide them.

PR #122, opened 2021-02-18
Fixes #31922 - Don't audit denials for rpm_script_tmp_t

ekohl:
https://bugzilla.redhat.com/show_bug.cgi?id=1651826 is the root cause for this. It was closed as WONTFIX for rh-ruby25 and RHEL8 so we'll likely see these for a while.

Having AVCs in audit logging that are "normal" means you can't easily distinguish real and "accepted". These AVCs should not be logged.

In fc3eb992a285bc701e8327815c43761a094b2a67 there was already a workaround for this but in the cron context. This commit removes those and changes it to the same dontaudit statements as for rpm_script_tmp_t.

This is an untested alternative to https://github.com/theforeman/foreman-selinux/pull/121.

### DIFF
--- a/foreman.te
+++ b/foreman.te
@@ -327,6 +327,21 @@ tunable_policy(`foreman_rails_can_connect_all',`
 rpm_exec(foreman_rails_t)
 rpm_read_db(foreman_rails_t)
 
+# Disregard denials for CVE-2019-3881 where bundler creates files in
+# /tmp/bundler if the home is not writable
+# This needs to be done for any type which can create /tmp/bundler
+gen_require(`
+    type rpm_script_tmp_t;
+    type system_cronjob_tmp_t;
+')
+
+dontaudit foreman_rails_t rpm_script_tmp_t:dir manage_dir_perms;
+dontaudit foreman_rails_t rpm_script_tmp_t:dir rw_dir_perms;
+dontaudit foreman_rails_t rpm_script_tmp_t:file manage_file_perms;
+dontaudit foreman_rails_t system_cronjob_tmp_t:dir manage_dir_perms;
+dontaudit foreman_rails_t system_cronjob_tmp_t:dir rw_dir_perms;
+dontaudit foreman_rails_t system_cronjob_tmp_t:file manage_file_perms;
+
 ######################################
 #
 # Logrotate and cron
@@ -334,7 +349,6 @@ rpm_read_db(foreman_rails_t)
 
 gen_require(`
     type logrotate_t;
-    type system_cronjob_tmp_t;
 ')
 
 # Allow sending signals in logrotate scripts
@@ -346,12 +360,6 @@ allow logrotate_t foreman_rails_unit_file_t:service manage_service_perms;
 # /etc/logrotate.d/foreman-proxy with logs as var_log_t
 files_list_var(logrotate_t)
 systemd_config_generic_services(logrotate_t)
-
-# When Foreman cronjob is started before Ruby on Rails, /tmp/bundler
-# is created with system_u:object_r:system_cronjob_tmp_t:s0 label denying
-# access to the web process
-manage_files_pattern(foreman_rails_t, system_cronjob_tmp_t, system_cronjob_tmp_t)
-manage_dirs_pattern(foreman_rails_t, system_cronjob_tmp_t, system_cronjob_tmp_t)
 
 #######################################
 #


lzap:
Don't we want to allow the action? Those `dontaudit` rules will prevent bundler from doing that and they will not be reported in the audit. I don't know why exactly bundler create these directories, but I assume it needs it?

Anyway, the fix that was merged was indeed not temporary, this is a behaviour that does not go anywhere. They just made the directory to be random: https://github.com/rubygems/bundler/pull/7416/files

ekohl:
My theory what happens: you run a rake task in RPM's %post. That creates /tmp/bundler with the RPM context. We really don't need that, but the next time you run Rails it wants to access `/tmp/bundler`. Another solution I thought about was using `PrivateTmp=true` in the systemd service so `/tmp` is completely private.

lzap:
Oh right, then let's simply change the `TMPDIR` variable bundler uses to pick the directory. It uses `Dir.tmpdir` which picks one of `[ENV['TMPDIR'], ENV['TMP'], ENV['TEMP'], @@systmpdir, '/tmp', '.']`. We would need to do this in all our scriplets, or maybe just `foreman-rake` could be enough.

foreman.te
```diff
@@ -327,14 +327,28 @@ tunable_policy(`foreman_rails_can_connect_all',`
...
+# /tmp/bundler if the home is not writable
+# This needs to be done for any type which can create /tmp/bundler
+gen_require(`
+    type rpm_script_tmp_t;
+    type system_cronjob_tmp_t;
+')
+
+dontaudit foreman_rails_t rpm_script_tmp_t:dir manage_dir_perms;
+dontaudit foreman_rails_t rpm_script_tmp_t:dir rw_dir_perms;
+dontaudit foreman_rails_t rpm_script_tmp_t:file manage_file_perms
```
evgeni:
```suggestion
dontaudit foreman_rails_t rpm_script_tmp_t:file manage_file_perms;
```

foreman.te
```diff
@@ -327,14 +327,28 @@ tunable_policy(`foreman_rails_can_connect_all',`
...
+dontaudit foreman_rails_t system_cronjob_tmp_t:dir manage_dir_perms;
+dontaudit foreman_rails_t system_cronjob_tmp_t:dir rw_dir_perms;
+dontaudit foreman_rails_t system_cronjob_tmp_t:file manage_file_perms
```
evgeni:
```suggestion
dontaudit foreman_rails_t system_cronjob_tmp_t:file manage_file_perms;
```

evgeni:
@lzap to the best of my understanding, we can just ignore the fact that bundler can't write to that folder, as we don't use bundler (in prod installs) to write things -- `bundle exec` is read-only, it just *tries* to create that folder.

I also don't exactly understand why the `cron` version was added in https://github.com/theforeman/foreman-selinux/commit/fc3eb992a285bc701e8327815c43761a094b2a67 -- seems unrelated to the commit message.

lzap:
Oh now I understand, so let's try with dontaudit then.

@ekohl the rule of thumb is - if a line starts with `allow` (or `dontaudit`) you need to end the line with a semicolon, otherwise it's a m4 macro and you must not put a semicolon there.

evgeni:
FWIW, with this change I can deploy a 2.4 box that has no AVC denials and rails seems to be happy (API, UI work)

evgeni (approved):

lzap:
Would you mind firing up some rake tasks like removing old reports? You can find these in cron jobs. Just to make sure.

evgeni:
Good point. When I fire it up *manually*, it's created as `user_tmp_t`, which also should be ignored I think.

evgeni:
Interesting, with `usr_tmp_t` label, I get:

```
Feb 22 09:51:26 centos7-foreman-2-4.yatsu.example.com foreman[7695]: /opt/rh/rh-ruby25/root/usr/share/gems/gems/bundler-1.16.1/lib/bundler.rb:193:in `rescue in tmp_home_path': `/usr/share/foreman` is not writable. (Errno::EEXIST)
Feb 22 09:51:26 centos7-foreman-2-4.yatsu.example.com foreman[7695]: Bundler also failed to create a temporary home directory at `/tmp/bundler/home':
Feb 22 09:51:26 centos7-foreman-2-4.yatsu.example.com foreman[7695]: File exists @ dir_s_mkdir - /tmp/bundler
```

when starting puma…

evgeni:
Ignoring rpm_script_tmp_t is fine, but cron and user seem to be breaking puma startup somehow. Weeeeeird.

evgeni:
~~But there are no SELinux denials, what.~~ of course, I ignore them

lzap:
Disable all dontaudit rules: `semodule -DB`, try again and then enable them back with `semodule -B`.

ekohl:
@evgeni now that we use `PrivateTmp=yes`, is this still needed or can we also drop the cron part?

evgeni:
I didn't test it, but I guess the current implementation is not required anymore.

I'd still expect user-invoked `foreman-rake` vs cron jobs to conflict, but that's not covered right now anyways

lzap:
I believe that private temp for httpd has no effect on cronjob, we either need to keep cronjob allowed or dontaudit it. I am against dropping it without further testing.

evgeni:
The current rule allows foreman_rails_t (httpd/puma) to touch cronjob_tmp_t, with PrivateTmp it will never even see the cronjob generated /tmp/bundler so IMHO it's safe to drop.

lzap:
For a moment, I thought that rake tasks are also running confined, it's not the case tho. So I agree then.

```
unconfined_u:unconfined_r:unconfined_t:s0-s0:c0.c1023 root 7240 0.0  0.0 185848 2392 pts/0 S+ 09:27   0:00 su foreman -s /bin/bash -c RUBYOPT=-W0 RAILS_ENV=production "$0" "$@" -- /usr/bin/tfm-rake reports:expire
```

PR #123, opened 2021-03-02
Fixes #31993 - add OpenStack documented ports

lzap:
Foreman can connect to OpenStack, however the API is fairly complicated with lots of endpoints and introspection. We have been adding ports on a "when it fails" basis but to improve experience, I would like to propose adding rules for all ports which are documented as a public API in OpenStack: https://access.redhat.com/documentation/en-us/red_hat_openstack_platform/16.1/html/firewall_rules_for_red_hat_openstack_platform/firewall_rules_for_red_hat_openstack_platform

See the RM issue for the grep command which reveals all those ports.

Also upstream OpenStack defines the following ports (some of them are covered by the base policy rules we already have) so I wonder - should we add the rest too: https://docs.openstack.org/install-guide/firewalls-default-ports.html

### DIFF
--- a/foreman-selinux-enable
+++ b/foreman-selinux-enable
@@ -10,7 +10,7 @@ LIBEXEC_DIR=/usr/libexec/foreman-selinux
 
 # Run hooks
 if [[ -d $LIBEXEC_DIR ]] ; then
-	find ${LIBEXEC_DIR} -name \*-before-disable.sh -type f -executable -exec /usr/bin/bash '{}' \;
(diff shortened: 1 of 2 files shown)


ekohl:
I'm not sure Foreman needs to connect to all services that OpenStack hsa. Do you plan to complete this or should we close it?

snecklifter:
RCA on this original issue was related to selinux denial on one of the sockets so you can safely close I think.

PR #125, opened 2021-03-05
Fixes #32023 - make ipa, abrt and logging macros optional

lzap:
There is the following denial on CO8Stream:

type=AVC msg=audit(1614873402.173:1566): avc: denied { getattr } for pid=30429 comm="httpd" path="/etc/puppetlabs/puppet/ssl/certs/centos8-stream-foreman-nightly.wisse.example.com.pem" dev="vda1" ino=33568393 scontext=system_u:system_r:httpd_t:s0 tcontext=unconfined_u:object_r:puppet_etc_t:s0 tclass=file permissive=0

This drills down to a problem with a different rule - ipa-selinux don't need to be installed by default and thus relevant macro will not expand correctly.


lzap:
For the record the other two macro types should be in optional macro too (their policies might not be present as well). Added that as a separate commit.

ekohl (commented):
Some additional context: in EL < 8.4, the ipa parts are in the base policy but on >= 8.4 they're in ipa-selinux. This can only be verified on Stream today. I'm leaning to merging this and waiting for a new package to verify. Unless @zpytela can point out some mistakes.

foreman.te
```diff
@@ -448,10 +454,12 @@ corenet_tcp_bind_vnc_port(websockify_t)
 corenet_tcp_bind_vnc_port(foreman_rails_t)
 dev_read_urand(websockify_t)
 kernel_read_system_state(websockify_t)
-logging_send_syslog_msg(websockify_t)
+optional_policy(`
+    logging_send_syslog_msg(websockify_t)
+    abrt_stream_connect(websockify_t)
```
zpytela, line 456: These modules are independent so should be in separate optional blocks.

foreman.te
```diff
@@ -593,6 +601,8 @@ optional_policy(`
     tunable_policy(`passenger_run_foreman', `
         files_manage_generic_tmp_files(passenger_t)
         files_manage_generic_tmp_dirs(passenger_t)
+        logging_search_logs(passenger_t)
```
zpytela, line 605: Even in tunable block it should be in a nested optional block.

zpytela:
Also note tftp is a module so all tftp_* interfaces calls should be in an optional block. The commit message should include puppet as they were also moved.

zpytela:
@ekohl actually ipa has their own module in RHEL 8 for some time already, the ipa module in selinux-policy was not updated so was outdated.
No bug mistake, the first commit is sufficient to prevent from the foreman-selinux subpackage installation.

foreman.te
lzap: It is. See line 600. I've learned that I can include tunable into optional but not vice versa, thus I moved it here instead.

foreman.te
lzap: Good find, fixed. Amended.

foreman.te
zpytela: Sorry, not enough context on my screen, exactly 1 line less.

zpytela:
//garbage removed
> Also note tftp is a module so all tftp_* interfaces calls should be in an optional block. The commit message should include puppet as they were also moved.

These comments are still valid.

Otherwise LGTM, but I only reviewed the commits and added what I recall from the previous discussion.

lzap:
Added TFTP as well, updated the commit message.

zpytela:
So puppet is now the only forgotten.

lzap:
Okay, made sure all puppet-related rules are in an optional block.

ekohl:
Looks like the build failed.

foreman.te
```diff
@@ -446,14 +454,20 @@ corenet_tcp_bind_vnc_port(websockify_t)
...
+optional_policy(`
+    abrt_stream_connect(websockify_t)
+')
 miscfiles_read_localization(websockify_t)
 sysnet_read_config(websockify_t)
-abrt_stream_connect(websockify_t)
 files_search_var_lib(websockify_t)
-read_files_pattern(websockify_t, puppet_var_lib_t, puppet_var_lib_t)
+optional_policy(`
+    read_files_pattern(websockify_t, puppet_var_lib_t, puppet_var_lib_t)
```
zpytela: It is not related to this PR, but puppet has `puppet_read_lib()`. It does not have a similar one for `puppet_etc_t` though anyway.

zpytela: Ok, there is `puppet_read_config()`, maybe too broad.

lzap: Do you have an idea why it fails? It looks like it's on this line but these errors are usually never the case.

lzap: Btw thanks I will refactor all of these, makes sense and makes it more cleaner, no need to require types explicitly now.

zpytela: I've never seen using patterns in an optional block, try to use the interfaces.

ekohl: websockify should only need to read the Puppet certificate files. In Puppet AIO those are nowadays `puppet_etc_t` since they live in `/etc` now instead of `/var/lib/puppet`. We don't support non-AIO on EL7 & EL8 so that should be safe to update.

lzap: Ok pushing.

foreman.te
```diff
@@ -591,14 +604,15 @@ optional_policy(`
...
+        logging_list_logs(passenger_t)
     ')
 ')
 optional_policy(`
     tunable_policy(`passenger_run_foreman', `
         manage_files_pattern(passenger_t, system_cronjob_tmp_t, system_cronjob_tmp_t)
         manage_dirs_pattern(passenger_t, system_cronjob_tmp_t, system_cronjob_tmp_t)
-        read_files_pattern(httpd_t, puppet_etc_t, puppet_etc_t)
-        read_lnk_files_pattern(httpd_t, puppet_etc_t, puppet_etc_t)
+        puppet_read_config(httpd_t)
```
ekohl, line 613: Isn't this duplicate from line 552 where it's unconditionally allowed?

lzap: Yes, we actually made copies of all rules into `passenger_run_foreman` blocks (for technical reason it can't be just a single block). This was done because we shipped versions with passenger and we were not sure about if we are fully dropping it.

Once you guys confirm that Passenger is gone for good, I will be more than happy remove all of these rules and the boolean flag as well. :-)

foreman.te
lzap: You were right it did not work previously, why? That's strange, I don't understand.

foreman.te
ekohl: I still don't understand why we effectively have:
```
 # Apache httpd reads puppet certs
optional_policy(`
    puppet_read_config(httpd_t)
') 

optional_policy(`
    tunable_policy(`passenger_run_foreman', `
        puppet_read_config(httpd_t)
    ')
')
```

Isn't this the same sufficient for both Puma and Passenger setups?
```
 # Apache httpd reads puppet certs
optional_policy(`
    puppet_read_config(httpd_t)
') 
```

ekohl:
@zpytela this is ok to be merged now?

zpytela:
I've just spotted
```
logging_search_logs(foreman_rails_t)
logging_list_logs(foreman_rails_t)
logging_search_logs(passenger_t)
logging_list_logs(passenger_t)
```
Only list is required as it is a superset of search. Otherwise LGTM.

lzap:
Ok removed `logging_search_logs` macros. Amended.

zpytela:
LGTM, no more comments.

ekohl (approved):
Thanks to you both!

tbrisker:
2.4 - 460a214 08939f7

tbrisker:
2.3 - b1e62b5 ce732e4

PR #128, opened 2021-05-31
Refs #32022 - Remove Ruby 1.8 paths

ekohl:
This removes the Ruby 1.8 paths as file contexts. EL6 shipped with Ruby 1.8 but that was dropped a long time ago. That means these files never show up anymore.

The issue isn't really resolved with it since restorecon still complains about missing files, but this reduces the output.

(cherry picked from commit 23a1a732c52d58dd8bbda8da59d519af09da4038)


lzap (approved):
Thanks.

PR #133, opened 2022-02-10
Replace CentOS Linux 8 in CI with CentOS Stream 8

ekohl:
foreman_docker is no longer a thing. This PR escalated from a fix to test with Stream 8 but in doing so I found more. See individual commits for details.

### DIFF
--- a/.github/workflows/compile.yaml
+++ b/.github/workflows/compile.yaml
@@ -13,7 +13,7 @@ jobs:
       matrix:
         centos: ["7", "8"]
     container:
-      image: centos:${{ matrix.centos }}
+      image: quay.io/centos/centos:${{ matrix.centos }}
 
     steps:
       - uses: actions/checkout@v2
--- a/.github/workflows/compile.yaml
+++ b/.github/workflows/compile.yaml
@@ -11,7 +11,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        centos: ["7", "8"]
+        centos: ["7", "stream8"]
     container:
       image: quay.io/centos/centos:${{ matrix.centos }}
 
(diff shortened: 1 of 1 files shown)


evgeni (approved):
APJ

ekohl:
Progress:
```
Makefile:19: *** *** Interface container.if or docker.if not present, cannot continue ***.  Stop.
```
Now I actually wonder why we have that. I suspect it's due to foreman_docker but that's dead. Thoughts about a patch to drop it?

evgeni (approved):

.github/workflows/compile.yaml
```diff
@@ -22,7 +22,7 @@ jobs:
         run: yum install -y selinux-policy-devel policycoreutils bzip2 perl
 
       - name: Compile policy
-        run: make DISTRO=rhel${{ matrix.centos }}
+        run: make DISTRO=rhel$(echo ${{ matrix.centos }} | sed 's/stream//')
```
evgeni: You could also do
```suggestion
        run: make DISTRO=rhel$(rpm --eval '%rhel')
```

Whether this is more readable or not, is up to you.

ekohl: I like the decoupling of the two. Implemented.

ekohl:
I couldn't make up my mind: I went back to just fixing CI since this will need to be cherry picked. The removal of old stuff will be a new PR.

evgeni (approved):

ekohl:
https://github.com/theforeman/foreman-selinux/pull/135 & https://github.com/theforeman/foreman-selinux/pull/136 for 3.2 and 3.1.

PR #138, opened 2022-04-06
Fixes #34730 - Drop docker/container integration

ekohl:
This used to be needed for foreman_docker but that has been dropped a while back. This means it can be removed from the policy.

It also drops some leftover type definitions.

### DIFF
--- a/foreman-selinux-enable
+++ b/foreman-selinux-enable
@@ -22,26 +22,12 @@ fi
 for selinuxvariant in targeted
 do
   if /usr/sbin/semodule -s $selinuxvariant -l >/dev/null; then
-    # Create port list cache
-    /usr/sbin/semanage port -E > $TMP_PORTS
-
-    # Remove previously defined conflicting docker_port_t (this can be removed in future release)
-    grep -E '(container|docker)_port_t' $TMP_PORTS | sed s/-a/-d/g >> $TMP_EXEC_BEFORE
-
-    # Commit changes
-    test -s $TMP_EXEC_BEFORE && /usr/sbin/semanage -S $selinuxvariant -i $TMP_EXEC_BEFORE
-
     # Load new policy
     /usr/sbin/semanage module -S $selinuxvariant -a /usr/share/selinux/${selinuxvariant}/foreman.pp.bz2
 
     # Create port list cache
     /usr/sbin/semanage port -E > $TMP_PORTS
 
-    # Assign foreman custom ports
-    grep -qE 'tcp 2375' $TMP_PORTS || \
-      echo "port -a -t foreman_container_port_t -p tcp 2375" >> $TMP_EXEC_AFTER
-    grep -qE 'tcp 2376' $TMP_PORTS || \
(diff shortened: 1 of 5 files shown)


ekohl:
@lzap do you know if elasticsearch and docker can be removed now:
https://github.com/theforeman/foreman-selinux/blob/10a124d3255af0d07f87d650b6a3c1ff31c9b80f/foreman-selinux-disable#L21-L27

lzap (approved):
Looks good, I haven't tested.

lzap:
> @lzap do you know if elasticsearch and docker can be removed now:

Yes, docker is removed in this patch, elastic was for Katello service and that was moved to Katello policy.

evgeni:
Elastic isn't used in Katello since uh, Satellite 6.3, whichever Katello version that was.

ekohl:
And katello-selinux dropped it in https://github.com/Katello/katello-selinux/commit/8cf33f7c7ea3169b417bfc0f67589477ed028143.

lzap:
Sure go ahead, I won't be merging as I do not want to be the one who breaks nightly. This should be safe, but I thought the same many times before :-)

foreman-selinux-enable
```diff
@@ -22,26 +22,12 @@ fi
 for selinuxvariant in targeted
 do
   if /usr/sbin/semodule -s $selinuxvariant -l >/dev/null; then
-    # Create port list cache
-    /usr/sbin/semanage port -E > $TMP_PORTS
-
-    # Remove previously defined conflicting docker_port_t (this can be removed in future release)
-    grep -E '(container|docker)_port_t' $TMP_PORTS | sed s/-a/-d/g >> $TMP_EXEC_BEFORE
```
lzap, line 29: One comment tho before you merge, removals in SELinux can be problematic when types are assigned to ports/files. You may get a compilation error on systems where foreman_container_port_t is still assigned to a port number. At least this was what was happening in older SELinux stacks.

This patch removes the type (see below) and it could have failed on SELinux policy reload, however, this grep/sed actually removes all types which end with container_port_t which includes foreman_container_port_t therefore it should not exist anymore and reload should be successful.

ekohl: I was also a bit worried about removal, but it isn't quite clear to me what you recommend here.

lzap: Sorry, I was saying out loud for myself - this is okay, let's merge this :-)

ekohl:
@evgeni / @ehelms do you mind taking a look and hit the green button if you agree we should clean this up?

ehelms (approved):

adamruzicka (approved):

adamruzicka:
Thank you @ekohl !

foreman-selinux-enable
evgeni: Narrator: it did, in fact, break upgrades: https://community.theforeman.org/t/luna-nightly-rpm-pipeline-354-failed/28201/3

PR #145, opened 2022-08-22
Fixes #35402 - label /var/lib/foreman/public httpd_sys_content_t

evgeni:
This is needed to allow it to serve static assets which are deployed to
/var/lib/foreman/public

### DIFF
--- a/foreman.fc
+++ b/foreman.fc
@@ -26,6 +26,7 @@
 /etc/puppetlabs/puppet/node.rb                  --  gen_context(system_u:object_r:foreman_enc_t, s0)
 
 /var/lib/foreman(/.*)?                              gen_context(system_u:object_r:foreman_lib_t,s0)
+/var/lib/foreman/public(/.*)?                       gen_context(system_u:object_r:httpd_sys_content_t,s0)
 
 /var/log/foreman(/.*)?                              gen_context(system_u:object_r:foreman_log_t,s0)
 


foreman.te
```diff
@@ -495,6 +495,9 @@ optional_policy(`
 files_search_pids(httpd_t)
 stream_connect_pattern(httpd_t, foreman_var_run_t, foreman_var_run_t, foreman_rails_t)
 
+# Allow Apache access to static assets
+allow httpd_t foreman_lib_t:file { getattr map open read };
```
adamruzicka: Afaik there are symlinks under /var/lib/foreman/webpack/. Do they need special handling here?

evgeni: They didn't in my testing.

`/usr/share/foreman/public` (the docroot of Apache) is a symlink to `/var/lib/foreman/public`, and all things in `/var/lib/foreman/public` are `foreman_lib_t`

If the question was "is there any special handling needed to resolve the symlink itself", I don't see this being needed (the symlink is `usr_t`, which Apache seems to be able to access)

ekohl: Shouldn't we label the public directory's contents as `httpd_sys_content_t` instead?

evgeni:
To test:
* Set `config.public_file_server.enabled` to `false` in `/usr/share/foreman/config/environments/production.rb` (see https://github.com/theforeman/foreman/pull/9365)
* Add `ProxyPass /webpack !` and `ProxyPass /assets !` to the `ProxyPass` list in `/etc/httpd/conf.d/05-foreman-ssl.conf` (CentOS) or `/etc/apache/sites-available/05-foreman-ssl.conf` (Debian) (see https://github.com/theforeman/puppet-foreman/pull/1078)
* Restart Apache and Foreman

foreman.te
evgeni: That's also possible (and has the same result in my setup).

~~Who would be responsible in doing the relabeling if we change it?~~

Packaging takes care of the relabeling too:
https://github.com/theforeman/foreman-packaging/blob/b38523b6dd1edb81d90e21875275b4ce97f29cff/packages/foreman/foreman-selinux/foreman-selinux.spec#L174-L177

ekohl: Is this line still needed? I'd imagine that it's (mostly) redundant.

evgeni: Nope, just me being lazy ;)

ekohl (approved):

PR #148, opened 2022-09-22
Install policy modules with priority 200

vmojzis:
https://fedoraproject.org/wiki/SELinux/IndependentPolicy#SELinux_Policy_module_priorities

Also, fix binary policy module path.

Signed-off-by: Vit Mojzis <vmojzis@redhat.com>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -87,8 +87,8 @@ install: install-policies \
 	install-data
 
 install-policies: policies consolidate-installation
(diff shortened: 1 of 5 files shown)


vmojzis:
Same changes should be applied to foreman-selinux.

ekohl (changes requested):
Thanks for the PR. Our automation requires a Redmine issue so I opened https://projects.theforeman.org/issues/35562. Could you amend your commit prefix the summary with `Fixes #35562 - `

PR #151, opened 2023-02-09
Fixes #36069 - remove old Puppet node.rb path

snagoor:
Even though `restorecon`  is passed with `-i` ignore switch, which only works if the path is valid. In newer versions of foreman, `/etc/puppet/` path is non-existent so is the error. 

To get of this error, removing the old Puppet path `/etc/puppet/node.rb`  from `foreman-selinux-relabel`  file.

### DIFF
--- a/foreman-selinux-relabel
+++ b/foreman-selinux-relabel
@@ -8,7 +8,6 @@
   /run/foreman.sock \
   /var/log/foreman \
   /etc/foreman \
-  /etc/puppet/node.rb \
   /etc/puppetlabs/puppet/node.rb \
   /etc/sysconfig/foreman* \
   /usr/lib/systemd/system/foreman* \


ekohl:
I still consider this a bug in `restorecon` and we should really submit it as a bug upstream. I'm not what the bugtracker for restorecon is though.

evgeni:
I guess bugzilla.redhat.com works ;)

ekohl:
I've closed the Redmine issue as a duplicate of https://projects.theforeman.org/issues/32022. Please use that.

And for the record, this happens when the parent directory doesn't exist:
```console
# restorecon /tmp/bla
restorecon: lstat(/tmp/bla) failed: No such file or directory
# restorecon -i /tmp/bla
# restorecon -i /tmp/path/bla
restorecon: SELinux: Could not get canonical path for /tmp/path/bla restorecon: No such file or directory.
```

ekohl:
I'm a bit hesitant to remove this line. Technically our installer can still put down the file in `/etc/puppet/node.rb`, though I'm not currently aware of any non-AIO puppetserver packaging effort.

So the good news is that https://bugzilla.redhat.com/show_bug.cgi?id=2137965 is actually already in VERIFIED and on CentOS Stream 8 I can see it does the right thing. The fix clearly didn't make it into Fedora 36, but I assume it's in newer Fedora versions.

This means that with RHEL 8.8 this will no longer be an issue, so I'm leaning to waiting for that to resolve it.

snagoor:
@ekohl thanks for the review. I have no problem in closing this PR as we already have a fix in the RHEL OS itself and if you think `/etc/puppet/node.rb` may have any additional value later as well. 

I assumed this file was no longer used, so thought of getting rid of it in the script itself.

ekohl:
> I assumed this file was no longer used, so thought of getting rid of it in the script itself.

It's in this weird place where there technically is a code path, but there really is a small chance to hit it. So I'm a bit torn on it, which is why I postponed acting on it for so long. Long enough for the fix to land in RHEL that I don't have to make a decision here.

ekohl:
Then I'm going to close it, but thanks for reminding me. I should have looked at reporting it upstream earlier, but I'm glad someone else did and it's getting fixed in the next RHEL release.

PR #163, opened 2024-02-02
make cockpit definition optional

evgeni:
The problem is that the cockpit policy moved to the cockpit package in EL9, not in the main selinux policy anymore.
Thus, when installing on a system without cockpit (and its policy), you get ugly errors like:

    Failed to resolve typeattributeset statement at /var/lib/selinux/targeted/tmp/modules/400/foreman/cil:66

And more importantly: the *whole* policy fails to load, leading to many issues at runtime.

To avoid this, we have to wrap the cockpit-related parts of the policy with `optional_policy` and introduce an own entrypoint that we can use in our file contexts, as there is no `optional()` for those.

### DIFF
--- a/foreman.fc
+++ b/foreman.fc
@@ -40,7 +40,7 @@
 
 # Foreman Remote Execution
 
-/usr/sbin/foreman-cockpit-session       gen_context(system_u:object_r:cockpit_session_exec_t,s0)
+/usr/sbin/foreman-cockpit-session       gen_context(system_u:object_r:foreman_cockpit_session_exec_t,s0)
 
 # Foreman Hooks plugin
 
--- a/foreman.te
+++ b/foreman.te
@@ -128,9 +128,6 @@ require{
     type bin_t;
     type httpd_t;
     type websm_port_t;
-    type cockpit_ws_t;
-    type cockpit_session_t;
-    type cockpit_session_exec_t;
     type unconfined_service_t;
     type http_cache_port_t;
     type squid_port_t;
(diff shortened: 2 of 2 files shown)


foreman.fc
```diff
@@ -40,8 +40,8 @@
 
 # Foreman Remote Execution
 
-/usr/sbin/foreman-cockpit-session       gen_context(system_u:object_r:cockpit_session_exec_t,s0)
-/usr/share/gems/gems/foreman_remote_execution-.*/extra/cockpit/foreman-cockpit-session -- gen_context(system_u:object_r:cockpit_session_exec_t,s0)
+#/usr/sbin/foreman-cockpit-session       gen_context(system_u:object_r:cockpit_session_exec_t,s0)
```
evgeni: This is obviously wrong, but I couldn't figure out how to produce an optional fcontext

foreman.te
```diff
@@ -333,25 +330,42 @@ manage_dirs_pattern(foreman_rails_t, system_cronjob_tmp_t, system_cronjob_tmp_t)
...
+optional_policy(`
+    gen_require(`
+        type cockpit_ws_t;
+        type cockpit_session_t;
+        type cockpit_session_exec_t;
+    ')
+
+    # foreman-cockpit-session needs to be labeled foreman_cockpit_session_exec_t,
+    # but we need to end up in cockpit_session_t
+    domtrans_pattern(cockpit_ws_t, foreman_cockpit_session_exec_t, cockpit_session_t)
```
evgeni, line 344: I need to check, this line might not be necessary, given the next one.

evgeni:
huh, this breaks fresh installs, wild:

```
error: lsetfilecon: (37 /usr/share/gems/gems/foreman_remote_execution-12.0.5/extra/cockpit/foreman-cockpit-session;65c4c92b, system_u:object_r:foreman_cockpit_session_exec_t:s0) Permission denied
error: Plugin selinux: hook fsm_file_prepare failed
```

evgeni:
```
# restorecon -v /usr/share/gems/gems/foreman_remote_execution-12.0.5/extra/cockpit/foreman-cockpit-session
restorecon: Could not set context for /usr/share/gems/gems/foreman_remote_execution-12.0.5/extra/cockpit/foreman-cockpit-session:  Permission denied

# dnf install cockpit
…

# restorecon -v /usr/share/gems/gems/foreman_remote_execution-12.0.5/extra/cockpit/foreman-cockpit-session
Relabeled /usr/share/gems/gems/foreman_remote_execution-12.0.5/extra/cockpit/foreman-cockpit-session from unconfined_u:object_r:usr_t:s0 to unconfined_u:object_r:foreman_cockpit_session_exec_t:s0
```

:exploding_head:

evgeni:
should be fixed by https://github.com/theforeman/foreman-packaging/pull/10404

evgeni:
I've moved the symlink-remove parts into https://github.com/theforeman/foreman-selinux/pull/164 as they are independent from EL9.

foreman.te
evgeni: it is required.

adamruzicka (approved):
To my limited understanding, this should be fine

PR #167, opened 2024-06-03
use AlmaLinux 8 for el8 tests

evgeni:
See: https://github.com/theforeman/foreman-infra/issues/2089

### DIFF
--- a/.github/workflows/compile.yaml
+++ b/.github/workflows/compile.yaml
@@ -11,11 +11,12 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        centos:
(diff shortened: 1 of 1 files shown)


ekohl (approved):
It would be nice if we could add a `name: ${{ matrix.container.split('/')[-1] }}` to improve readability, but I don't think it's that easy.

evgeni:
> It would be nice if we could add a `name: ${{ matrix.container.split('/')[-1] }}` to improve readability, but I don't think it's that easy.

Yeah, GHA doesn't know how to mangle strings :(

PR #168, opened 2024-07-26
Fixes #37999 - allow smart-proxy with PuppetCA to read some etc files

PopiBrossard:
Dear maintainer,

The current foreman-proxy SELinux policy isn't working when trying to use the PuppetCA feature. The proxy tries to read some files in `/etc/foreman-proxy` but is not allowed:

```
type=AVC msg=audit(1721979897.417:100790): avc:  denied  { read } for  pid=731469 comm="smart-proxy" name="puppetca_hostname_whitelisting.yml" dev="dm-0" ino=33791767 scontext=system_u:system_r:foreman_proxy_t:s0 tcontext=system_u:object_r:hostname_etc_t:s0 tclass=file permissive=1
type=AVC msg=audit(1721979897.417:100790): avc:  denied  { open } for  pid=731469 comm="smart-proxy" path="/etc/foreman-proxy/settings.d/puppetca_hostname_whitelisting.yml" dev="dm-0" ino=33791767 scontext=system_u:system_r:foreman_proxy_t:s0 tcontext=system_u:object_r:hostname_etc_t:s0 tclass=file permissive=1
type=AVC msg=audit(1721979897.417:100791): avc:  denied  { ioctl } for  pid=731469 comm="smart-proxy" path="/etc/foreman-proxy/settings.d/puppetca_hostname_whitelisting.yml" dev="dm-0" ino=33791767 ioctlcmd=0x5401 scontext=system_u:system_r:foreman_proxy_t:s0 tcontext=system_u:object_r:hostname_etc_t:s0 tclass=file permissive=1
type=AVC msg=audit(1721979897.417:100792): avc:  denied  { getattr } for  pid=731469 comm="smart-proxy" path="/etc/foreman-proxy/settings.d/puppetca_hostname_whitelisting.yml" dev="dm-0" ino=33791767 scontext=system_u:system_r:foreman_proxy_t:s0 tcontext=system_u:object_r:hostname_etc_t:s0 tclass=file permissive=1
```

To make this work, I've created a new file type `foreman_proxy_etc_t` and added the minimum privileges.

Regards

PR #172, opened 2024-09-26
Remove dependency on unconfined selinux module

matthewdva:
Moved the temporary rule to prevent dontaudit denies during startup to an optional policy.  This change removed the dependence on the unconfined selinux module.

### DIFF
--- a/foreman.te
+++ b/foreman.te
@@ -128,7 +128,6 @@ require{
     type bin_t;
     type httpd_t;
     type websm_port_t;
-    type unconfined_service_t;
     type http_cache_port_t;
     type squid_port_t;
 }
@@ -144,7 +143,12 @@ init_daemon_domain(foreman_rails_t, foreman_rails_exec_t)
 # Temporary rule to prevent dontaudit denial during start caused
 # by socket activation:
 # https://community.theforeman.org/t/foreman-nightly-rpm-pipeline-611-failed/19179
-allow foreman_rails_t unconfined_service_t:tcp_socket { connected_stream_socket_perms };
+optional_policy(`
+    require{
+        type unconfined_service_t;
+    }
+    allow foreman_rails_t unconfined_service_t:tcp_socket { connected_stream_socket_perms };
+')
 
 # Socket and PID files transition
 files_pid_filetrans(foreman_rails_t, foreman_var_run_t, { file dir sock_file })


ekohl (commented):
Apologies for the late response.

Looking at the thread, the Smart Proxy service ends up running unconfined and I'd say that in most setups we have that running on the same machine. Can you tell me why you don't want to depend on the unconfined service?

Today we don't really support the `foreman-proxy` SELinux module and nobody is working on completing it so I'm not sure we can improve this.

matthewdva:
As part of our security policy, we disable the unconfined module on all systems. Because the policy is dependent on the unconfined module, we are unable to install it on the server or a smart proxy.

When we encounter applications that do not work correctly without the unconfined module, we will create auxiliary modules to correct any issues.  We prefer to leverage the upstream modules that are provided when they exist.   We cannot leverage the foreman-selinux without patching the policy and rebuilding the rpm. 

The comment regarding the dependency on the unconfined modules dependency states it is a temporary policy to prevent excess logging.  We are proposing this temporary policy become optional for those like us who want to disable the unconfined module.

ekohl:
> As part of our security policy, we disable the unconfined module on all systems. Because the policy is dependent on the unconfined module, we are unable to install it on the server or a smart proxy.

That helps to understand this.

> When we encounter applications that do not work correctly without the unconfined module, we will create auxiliary modules to correct any issues. We prefer to leverage the upstream modules that are provided when they exist. We cannot leverage the foreman-selinux without patching the policy and rebuilding the rpm.

I'm happy to merge patches to address this. Today it's not supported because it's not finished (https://github.com/theforeman/foreman-selinux/pull/168 sort of proves that).

> The comment regarding the dependency on the unconfined modules dependency states it is a temporary policy to prevent excess logging. We are proposing this temporary policy become optional for those like us who want to disable the unconfined module.

That's fine by me. And I suspect that if you do run foreman-proxy within a policy you may not need to connect to unconfined TCP sockets, though I don't think there's a policy for puppetserver either which could shift the problem where foreman-proxy connects to an unconfined service.

To move this forward, could you please open a Redmine issue on https://projects.theforeman.org/projects/selinux and link it here?

matthewdva:
> To move this forward, could you please open a Redmine issue on https://projects.theforeman.org/projects/selinux and link it here?
I have been trying to get a Redmine account so I can create an issue.  I keep getting a message, "Your account was created and is now pending administrator approval."  I am happy to create an issues once I can sign in.

ekohl:
I'd have expected to send an email, but I've now activated your user.

evgeni:
I thought I had it configured not to require activation at all if you login via GitHub 🤔

matthewdva:
Following up to see if anything else is needed to move this request forward.

ekohl (approved):
I just kicked off CI. @evgeni any objections?

evgeni (approved):

ekohl:
Thanks @matthewdva!

PR #175, opened 2024-11-19
Test on CentOS Stream 10

ekohl:
Should this also deal with Packit or is that too early?


evgeni:
> Should this also deal with Packit or is that too early?

There is no RHEL10 chroot yet. We could use `centos-stream-10-x86_64`, sure.

ekohl:
`foreman-selinux-3.14.0-0.20241206125424585417.pr175.el10.noarch`: it compiles, ship it?

evgeni (approved):

evgeni:
:ship:

PR #177, opened 2025-01-31
label 24443 as IOP Advisor Engine port and allow Foreman to connect

evgeni:


ekohl (commented):
Since foreman_rh_cloud depends on Katello, does this really belong in katello-selinux?

evgeni:
> Since foreman_rh_cloud depends on Katello, does this really belong in katello-selinux?

Maybe. Or maybe we should put `katello-selinux` in here and treat it as a plugin? :)

(read: I don't strictly care where it ends up)

ekohl:
I don't mind merging it, but practically: katello-selinux depends on candlepin-selinux which complicates it. For now I'd submit it in katello-selinux

evgeni:
https://github.com/theforeman/katello-selinux/pull/59

PR #179, opened 2025-07-15
Fixes #38583: Allow mounting postgres as socket to container

ehelms:
This is part of the IoP work where the containers are mounting Postgres as a socket. The denials:

```
type=AVC msg=audit(1752541397.045:4832): avc:  denied  { write } for  pid=49619 comm="python3.12" name=".s.PGSQL.5432" dev="tmpfs" ino=1528 scontext=system_u:system_r:container_t:s0:c605,c856 tcontext=system_u:object_r:postgresql_var_run_t:s0 tclass=sock_file permissive=1
type=AVC msg=audit(1752541397.045:4832): avc:  denied  { connectto } for  pid=49619 comm="python3.12" path="/run/postgresql/.s.PGSQL.5432" scontext=system_u:system_r:container_t:s0:c605,c856 tcontext=system_u:system_r:postgresql_t:s0 tclass=unix_stream_socket permissive=1
```

### DIFF
--- a/foreman.te
+++ b/foreman.te
@@ -274,6 +274,14 @@ optional_policy(`
     postgresql_stream_connect(foreman_rails_t)
 ')
 
+# Allow containers to connect to PostgreSQL
+optional_policy(`
+    gen_require(`
+        type container_t;
+    ')
+    postgresql_stream_connect(container_t)
+')
+
 # Connecting to Redis
 corenet_tcp_connect_redis_port(foreman_rails_t)
 


adamruzicka (approved):
To be honest I'm not fully up to speed with how selinux and containers play together, but it does the trick on my box.

adamruzicka:
And apparently linking to redmine is a hard requirement here

foreman.te
```diff
@@ -274,6 +274,15 @@ optional_policy(`
     postgresql_stream_connect(foreman_rails_t)
 ')
 
+# Allow containers to connect to PostgreSQL
+optional_policy(`
+    gen_require(`
+        type container_t;
+    ')
+    allow container_t postgresql_var_run_t:sock_file write;
```
evgeni: This should be covered by `postgresql_stream_connect(container_t)`.

(`postgresql_stream_connect` includes `stream_connect_pattern`, which in turn includes `allow …:sock_file write_sock_file_perms;` and `write_sock_file_perms` is `{ getattr write open append }`

evgeni (changes requested):
Please add a redmine and check that the explicit `allow` is not needed

(ACK on the change as such)

adamruzicka:
Still works, still misses redmine

evgeni (approved):

PR #181, opened 2025-07-24
Allow foreman_rails_t to domtrans to ssh_t

evgeni:


lzap:
There is no reason to transition to `ssh_t`, in fact it makes things more complex. Our domain already contains everything that SSH client needs, the rest is just `sshpass`: https://github.com/theforeman/foreman-selinux/pull/187

PR #184, opened 2026-01-12
Allow Foreman to connect to OpenShift for KubeVirt

evgeni:


foreman.te
```diff
@@ -542,3 +542,22 @@ require {
...
 allow foreman_rails_t tmpfs_t:filesystem getattr;
 allow foreman_rails_t fs_t:filesystem getattr;
+
+######################################
+#
+# Foreman KubeVirt plugin
+#
+
+type foreman_openshift_port_t;
+corenet_port(foreman_openshift_port_t)
```
evgeni: bikeshedding question: We (Foreman) need it for the KubeVirt integration, but KubeVirt is just a plugin to OpenShift and `6443` is the API port of OpenShift, so I named it `foreman_openshift_port_t`. Are y'all OK with that?

foreman-selinux-disable
```diff
@@ -19,7 +19,7 @@ if /usr/sbin/semodule -s $selinuxvariant -l >/dev/null; then
 
   # Remove all user defined ports (including the default one)
   /usr/sbin/semanage port -E | \
-    grep 'foreman_proxy_port_t' | \
+    grep 'foreman_.*_port_t' | \
```
ekohl, line 22: I guess our disable script never really worked well and likely that nobody used it for a long time.

ekohl (approved):
I'm wondering if we need to document changing the OpenShift port. Does OpenShift even allow that?

foreman-selinux-disable
evgeni: same

evgeni:
> I'm wondering if we need to document changing the OpenShift port. Does OpenShift even allow that?

To the best of my knowledge OpenShift doesn't allow that.

https://access.redhat.com/solutions/4665121

ekohl:
Can you open a Redmine issue so we can merge this?

evgeni:
Packit seems to be down right now, but here is a manual build if you want to try things out: https://download.copr.fedorainfracloud.org/results/evgeni/packit-scratch/rhel-9-x86_64/09999865-foreman-selinux/

evgeni:
(and packit is back)

foreman-selinux-disable
Gauravtalreja1: Just curious, when do we run disable script and how this could be validated?

evgeni: on uninstallation of the package, for example (but it also can be ran manually, to break the system :see_no_evil: )

Gauravtalreja1: which package you mean?

evgeni: foreman-selinux

Gauravtalreja1: ah, so when I try to remove foreman-selinux, it tries to remove protected packages like satellite, so not sure how we could test the disable script?

evgeni: ah yes, of course :see_no_evil: 

you could test it by downgrading the package to the one in nightly, which will also execute it (as upgrade/downgrade is just a special case of "uninstall" and "install") and see that there are no errors and afterwards the port is not listed in `semanage port -l` anymore

foreman.te
ShimShtein: At the risk of sounding too strict: `foreman_openshift_api_port_t`

evgeni: Nah, that's a good idea!

evgeni: Updated

foreman-selinux-disable
Gauravtalreja1: @evgeni had tested this on an empty centos box (without Foreman), by just installing foreman-selinux and then removing it again, and it works correctly.

Gauravtalreja1 (approved):
ACK, I’ve tested the changes and they work as expected 🍏

PR #189, opened 2026-09-22
Fixes #39790 - Label generic Foreman worker entry point

jakduch:
Labels the new `/usr/libexec/foreman/foreman-worker` entry point as `foreman_rails_exec_t`, preserving the same SELinux domain transition used by the existing `*-selinux` wrapper.

This accompanies theforeman/foreman#11275. RPM installation is handled separately, and theforeman/foremanctl#429 will consume the entry point.

AI assistance disclosure:
- Assisted by Codex 5.6 Sol High.


jakduch:
This is the SELinux follow-up to theforeman/foreman#11275 and Redmine #39789. The Redmine check requires a dedicated ticket in the SELinux project, so I’ll retarget the commit once that follow-up ticket is created.

jakduch:
Updated to the dedicated SELinux ticket #39790.

